PR #7, opened 2019-01-29
Add missing test artifacts on linux

ziyu-guo:
Add missing artifacts and data files for integration tests on linux platforms.


wweic:
Thanks @ziyu-guo . Could you also add documentation about how to run these tests?

zhiics:
I think adding a few artifacts like this is okay, but we probably need to figure out a better way to add artifacts for integration tests because the the params files are usually quite big. Hence, the size of test folder will blow if we keep adding more and more integration test.

wweic:
Agree @zhiics. Probably we can store the integration test artifacts in s3 and download on demand.

wweic:
created https://github.com/neo-ai/neo-ai-dlr/issues/9 to follow up the integrate test artifacts issue.

PR #18, opened 2019-02-07
Add build doc

hcho3:
Also add options to CMake so that users can choose to enable CUDA from command line.

TODO. Revise CMakeLists.txt to enable CUDA compilation on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,11 @@ include(cmake/Utils.cmake)
 set_default_configuration_release()
 msvc_use_static_runtime()
 
+# Options
+option(USE_CUDA  "Build with CUDA" OFF)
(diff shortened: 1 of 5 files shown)


ziyu-guo:
Do we want to add `--user` to the python install cmdlines?

hcho3:
@ziyu-guo `--user` is only available from Pip. Should we tell the user to build wheels first? The commands would be

```bash
cd python
python3 setup.py bdist_wheel
pip3 install --user dist/*.whl
```

ziyu-guo:
@hcho3 I mean like on [this link](https://docs.tvm.ai/install/from_source.html#python-package-installation)

`cd python; python setup.py install --user; cd ..`
`cd topi/python; python setup.py install --user; cd ../..`
`cd nnvm/python; python setup.py install --user; cd ../..`

PR #22, opened 2019-02-13
Add OpenCL support for Deeplens

ziyu-guo:
Adding OpenCL support for devices like deeplens, and adjust cuda lib location overriding logic.


CMakeLists.txt
```diff
@@ -61,9 +62,35 @@ FILE(GLOB_RECURSE DLR_INC
     "${TREELITE_SRC}/include/*.h"
 )
 
+if(USE_OPENCL)
+    message("USING OpenCL")
+    if(USE_OPENCL STREQUAL "ON")
+      set(USE_OPENCL "/opt/intel/opencl")
```
zhiics: Should we have an (environment) variable for the paths instead of hard code them here?

CMakeLists.txt
```diff
@@ -61,9 +62,35 @@ FILE(GLOB_RECURSE DLR_INC
...
+    list(APPEND TVM_RUNTIME_LINKER_LIBS ${OPENCL_LIBRARY})
+    file(GLOB RUNTIME_OPENCL_SRCS ${TVM_SRC}/src/runtime/opencl/*.cc)
+    list(APPEND DLR_SRC ${RUNTIME_OPENCL_SRCS})
+endif()
+
 if(USE_CUDA)
     message("USING CUDA")
-    set(USE_CUDA "/usr/local/cuda-9.0")
+    if(USE_CUDA STREQUAL "ON")
+      set(USE_CUDA "/usr/local/cuda-9.0")
```
yongwww: would be better to not hard code cuda-9.0, considering cuda 10 was released

CMakeLists.txt
ziyu-guo: Fixed in new commit, can you check?

CMakeLists.txt
ziyu-guo: Fixed in new commit, can you check?

zhiics:
@ziyu-guo LGTM

CMakeLists.txt
```diff
@@ -61,9 +63,32 @@ FILE(GLOB_RECURSE DLR_INC
...
+      set(OpenCL_TOOLKIT_ROOT_DIR ${USE_OPENCL})
+      message(STATUS "Custom OPENCL_PATH=" ${OpenCL_TOOLKIT_ROOT_DIR})
+      set(OpenCL_INCLUDE_DIRS ${OpenCL_TOOLKIT_ROOT_DIR}/include)
+      set(OpenCL_LIBRARIES ${USE_OPENCL}/lib/libOpenCL.so)
+    endif(USE_OPENCL STREQUAL "ON")
+    include_directories(${OpenCL_INCLUDE_DIRS})
+    list(APPEND TVM_RUNTIME_LINKER_LIBS ${OpenCL_LIBRARIES})
+    file(GLOB RUNTIME_OPENCL_SRCS ${TVM_SRC}/src/runtime/opencl/*.cc)
+    list(APPEND DLR_SRC ${RUNTIME_OPENCL_SRCS})
+    set(USE_OPENCL OFF)
```
hcho3: Why do we turn off `USE_OPENCL` at the end?

ziyu-guo: There seems to be a bug in dmlc/tvm's OpenCL build. Plus DLR OpenCLbuild does not rely on tvm runtime being builting with OpenCL, so turning it off to decouple the 2 USE_OPENCL switches (same name).

PR #26, opened 2019-02-21
Fix issue in output size calculation

ziyu-guo:
Fixes output size calculation when model has multiple outputs.

Also  make sure output_size_dim is filled in correct order.


python/dlr/api.py
```diff
@@ -220,7 +220,10 @@ def _get_output_shape(self, index):
             The shape of the index-th output.
         """
         size, dim = self._get_output_size_dim(index)
-        self.output_size_dim.append((size, dim))
+        if index >= len(self.output_size_dim):
```
zhiics: I don't quite understand here. Why index would overflow?

python/dlr/api.py
```diff
@@ -220,7 +220,10 @@ def _get_output_shape(self, index):
...
+            self.output_size_dim.extend( [(0, 0)] * (index - len(self.output_size_dim) + 1))
+        print( self.output_size_dim)
```
zhiics: Remove print

python/dlr/api.py
ziyu-guo: In old code, self.output_size_dim would be in correct order only if _get_output_shape(index) gets called with index in consecutive order (0, 1, 2 etc). Furthermore, repeated _get_output_shape() calls will write redundant entries to this list.

zhiics: Discussed offline. No need to have the logic to check and extend the size of the output array. Just resizing to the real number of outputs and then indexing directly should work.

zhiics:
LGTM, but probably we need to add a unit test. This way DLR would gradually have better test coverage.

tests/python/integration/4in2out/multi_input.py
```diff
@@ -14,24 +14,35 @@
...
-params["data3"] = data3 = np.random.uniform(-1, 1, size=data_shape).astype("float32")
-params["data4"] = data4 = np.random.uniform(-1, 1, size=data_shape).astype("float32")
+shape_dict.update({"data3": data_shape2, "data4": data_shape2})
+params["data3"] = data3 = np.random.uniform(-1, 1, size=data_shape2).astype("float32")
+params["data4"] = data4 = np.random.uniform(-1, 1, size=data_shape2).astype("float32")
 
 net = symbol.Group([net1, net2])
 
+
+print('=======')
```
zhiics: remove these prints. no need to print, assertions are just fine.

PR #30, opened 2019-02-28
Add amd64 CUDA target in automatic build pipeline

hcho3:
The automatic build will use Ubuntu 14.04, CUDA 8.0, CuDNN 7, and TensorRT 4.


hcho3:
Closing for now, will try to run more experiments locally until things work out.

PR #32, opened 2019-03-13
Add install test for cloud targets

hcho3:
Test whether the x86 DLR wheel works on C4, C5, M4, M5 cloud targets.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,11 @@
 // Jenkins pipeline
 // See documents at https://jenkins.io/doc/book/pipeline/jenkinsfile/
 
+// List of cloud targets
+def cloudTargetMatrix = [
+  "c4", "c5", "m4", "m5"
+]
+
 /* Pipeline definition */
 pipeline {
   // Each stage specify its own agent
@@ -33,6 +38,15 @@ pipeline {
         script {
           parallel ([ "build-amd64-cpu" : { AMD64BuildCPU() },
                       "build-amd64-gpu" : { AMD64BuildGPU() } ])
+        }
+      }
+    }
+    stage('Jenkins: Install & Test') {
+      steps {
+        script {
+          parallel (cloudTargetMatrix.collectEntries{
+            [(it): { CloudInstallAndTest(it) } ]
+          })
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/dlr.cc
```diff
@@ -184,14 +184,12 @@ void DLRModel::SetupTVMModule(const std::string& model_path) {
 void DLRModel::SetupTreeliteModule(const std::string& model_path) {
   ModelPath paths = get_treelite_paths(model_path);
   int num_worker_threads = -1; // use the maximum amount of threads
-  int include_master_thread = 1;
```
ziyu-guo: What is it for?

hcho3: This corresponds to the API change in the upstream. The `include_master_thread` parameter was causing confusion and thus removed in dmlc/treelite#79

Jenkinsfile
```diff
@@ -88,3 +102,21 @@ def AMD64BuildGPU() {
     }
   }
 }
+
+def CloudInstallAndTest(cloudTarget) {
+  def nodeReq = "ubuntu && amd64 && ${cloudTarget}"
+  node(nodeReq) {
+    echo "Installing DLR package for ${cloudTarget} target"
+    withAWS(credentials:'Neo-AI-CI-Fleet') {
+      files = s3FindFiles(bucket: 'neo-ai-dlr-jenkins-artifacts', glob: "${env.JOB_NAME}/${env.BUILD_ID}/artifacts/dlr-*-manylinux1_x86_64.whl")
```
wweic: How are these files in `neo-ai-dlr-jenkins-artifacts` generated?

hcho3: In the Build stage

Jenkinsfile
```diff
@@ -3,6 +3,11 @@
 // Jenkins pipeline
 // See documents at https://jenkins.io/doc/book/pipeline/jenkinsfile/
 
+// List of cloud targets
+def cloudTargetMatrix = [
+  "c4", "c5", "m4", "m5"
```
wweic: why not also add `p2` `p3`?

Jenkinsfile
wweic: Got it. Thank you!

Jenkinsfile
hcho3: Let’s wait until the EC2 instance limit is increased?

hcho3:
I now realize that history re-writing invalidates submodule links in previous history. From now on, I will do periodic merge with upstream, not rebase

Jenkinsfile
wweic: Make sense.

PR #38, opened 2019-03-22
Fix segfault for multi-class classifier trees with max class output

hcho3:
Currently, DLR crashes with segmentation fault if the multi-class classifier tree model from [this notebook](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/introduction_to_amazon_algorithms/xgboost_mnist/xgboost_mnist.ipynb) is loaded into DLR.

**Diagnosis**. The MNIST model produces a single number to indicate the most probable class output, so DLR only allocates a single floating-point slot for the output. This is wrong, since intermediate calculations require 10 floating-point slots, not one (10 is the number of classes).

**Fix**. Allocate sufficient space for intermediate calculations; the amount is indicated with `treelite_output_buffer_size_`. Also, use `std::vector<float>` instead of `std::unique_ptr<float[]>` to avoid unnecessary memory re-allocation.


wweic:
@hcho3 could you add a unit test?

PR #41, opened 2019-03-29
Retrieve matching linux TVM test model artifacts from S3 based on Processor Architecture

### LINKED ISSUE #9: Store integration test artifacts in S3 and download on demand
so we won't need to store them in the repo.

ziyu-guo:
And remove model artifacts from this repo

Public S3 folder at s3://neo-ai-dlr-test-artifacts/

Closes #9


hcho3:
Let me also add Treelite artifacts too

hcho3:
@ziyu-guo For "x86_64" target, which instance did you use to compile? Models compiled for C5 target doesn't run on C4 or M4, for example (due to use of AVX512).

ziyu-guo:
> @ziyu-guo For "x86_64" target, which instance did you use to compile? Models compiled for C5 target doesn't run on C4 or M4, for example (due to use of AVX512).

X64_64 artifacts are compiled with just "-llvm" in centos6 docker for compatibility. These are CPU-only models.

hcho3:
@ziyu-guo So it's safe to assume that it would run for all x86_64 targets (it won't use fancy AVX etc)?

ziyu-guo:
@hcho3 Yes, compatibility instead of opt level in this correctness test.

wweic:
Change LGTM. Could you also add the integration test into Jenkins? @ziyu-guo

hcho3:
@wweic Will add the test into Jenkins today.

hcho3:
Uh oh, the assertion for LETOR is failing. Let me run this on my machine.

PR #45, opened 2019-04-13
Add android on ARM support

ziyu-guo:
*Description of changes:*

Add android build support and instructions


wweic:
Is it possible to write a script that builds DLR using NDK? You can find some docker images with NDK installed in [DockerHub](https://hub.docker.com/search/?q=android-ndk&type=image). We should make that part of our Jenkins build.

ziyu-guo:
@wweic This change uses NDK standalone tool chain under DLR's own cmake. The only difference is that we use NDK-provided headers and libs. 

Regarding jenkins, we probably need a separate container for Android build since  I remember glibc 2.14 is required for NDK-provided executables. Will sync up with @hcho3 to see how we can set it up.

Issue #46, opened 2019-04-19
[RFC] Support alternative runtimes


apivovarov:
I'd like to discuss how DLR can support alternative runtimes. e.g. Tensorflow, Tensorflow Lite, etc.

Currently we have `DLRModel` class which accepts `model_path` and `device` parameters. 
We can use model files format to determine what runtime should be used. E.g. if model is represented by `.pb` file then it indicates that DLR needs to use `tensorflow` runtime.

In order to support for Tensorflow runtime and at the same time preserve current API we can do the following refactoring:
- add IDLRModel abstract class with `@abc.abstractmethod run(self, input_data)` and other "public" method
- move all DLRModel implementation to DLRModelImpl class
- make DLRModelImpl to be sub-class of IDLRModel
- add `TFModelImpl(IDLRModel)` class which incapsulates details on how to run tensorflow models.
- make DLRModel to be a wrapper on top of particular implementation. __init__ method will decide which impl to create based on format of the model files.

The diagram and the code below illustrate the idea
### Diagram
![DLRModel](https://user-images.githubusercontent.com/218087/56775105-6930fd00-677a-11e9-82fe-913ea0903af8.png)

### Code Example
```
import abc

# Interface
class IDLRModel:
    __metaclass__=abc.ABCMeta

    @abc.abstractmethod
    def get_input_names(self):
        return


    @abc.abstractmethod
    def get_input(self, name, shape=None):
        return


    @abc.abstractmethod
    def run(self, input_data):
        return

# Wrapper class
class DLRModel(IDLRModel):
    def __init__(self, model_path, device):
        if model_path.endswith(".pb"):
            self._impl = TFModelImpl(model_path, device)
        else:
            self._impl = DLRModelImpl(model_path, device) 


    def run(self, input_data):
        return self._impl.run(input_data)


# Current DLRModel code will be moved to  DLRModelImpl
class DLRModelImpl(IDLRModel):
    def __init__(self, model_path, device):
        self.model_path = model_path
        self.device = device


    def run(self, input_data):
        return "TVM/Treelite model run {}".format(input_data)


# New class which support .pb models execution on tensorflow runtime
class TFModelImpl(IDLRModel):
    def __init__(self, model_path, device):
        self.model_path = model_path
        self.device = device


    def run(self, input_data):
        # import tensorflow as tf
        # res = sess.run(output_tensors, feed_dict={input_tensor: np_input})
        # return res
        return "TF model run {}".format(input_data)
```
[...]

PR #47, opened 2019-04-24
Add Tensorflow Model Impl

apivovarov:
*RFC #46*

## Description of changes:
TFModelImpl is a wrapper on top of tensorflow which implements DLRModel API.
It allows to load frozen_graph_file and run inference

## Next Step
This PR is the first PR to implement proposal #46. This PR focuses only on `TFModelImpl` itself.
The next PR will refactor existing `DLRModel` into `IDLRModel` interface, `DLRModel` dispatcher and `DLRModelImpl`.
Once it is done we can integrate `TFModelImpl` into new architecture.
1. make `TFModelImpl` inherit from `IDLRModel` (to make sure it implements `run()`, `get_input_names()`, etc methods)
2. add code to `DLRModel` dispatcher which recognize TF frozen graph file (.pb file)
3. if frozen graph file is found `DLRModel` will create `TFModelImpl` instance (`self._impl`) and will use it during `DLRModel` methods invocations by user.

The diagram below illustrates the idea
### Diagram
DLRModel

### DIFF
--- a/python/dlr/tf_model.py
+++ b/python/dlr/tf_model.py
@@ -1,0 +1,181 @@
+# coding: utf-8
+import logging
+import os
+import tensorflow as tf
+
+# A prefix that will be prepended to the names in graph_def
+PREFIX = "import"
+
+
+def _load_frozen_graph(frozen_graph_file, device):
+    """
+    Load Frozen Graph
+
+    Parameters
+    ----------
+    frozen_graph_file : str
+        Full path to frozen graph (.pb file)
+    device : str
+        device type and id, (e.g. /cpu:0)
+
+    Returns
+    -------
+    out : Graph :py:class:`tf.Graph`
+    """
+    logging.info("Loading frozen graph: {}".format(frozen_graph_file))
+    with tf.device(device):
+        with tf.gfile.GFile(frozen_graph_file, 'rb') as f:
+            graph_def = tf.GraphDef()
+            graph_def.ParseFromString(f.read())
+        with tf.Graph().as_default() as graph:
+            tf.import_graph_def(graph_def, name=PREFIX)
+        return graph
+
+
+def _get_input_and_output_names(graph):
+    """
+    Get input and output tensor names
+
+    Parameters
+    ----------
+    graph : tf.Graph
+        Tensorflow graph
+
+    Returns
+    -------
+    input_tensor_names : List of input tensor names
+    output_tensor_names  : List of output tensor names
+    """
+    input_tensor_names = []
+    output_tensor_names = set()
+    op_prefix = PREFIX + "/"
+    for op in graph.get_operations():
+        if not op.name.startswith(op_prefix):
+            continue
+        if op.type == 'Placeholder' and op.inputs.__len__() == 0 and op.outputs.__len__() == 1:
+            input_tensor_names.append(op.outputs[0].name)
+        if op.outputs.__len__() == 1:
+            output_tensor_names.add(op.outputs[0].name)
+    for op in graph.get_operations():
+        for in_t in op.inputs:
+            if in_t.name in output_tensor_names:
+                output_tensor_names.remove(in_t.name)
+    # Sort list of output tensor names in order to get consistent output in run()
+    output_tensor_names = list(output_tensor_names)
+    output_tensor_names.sort()
+    return input_tensor_names, output_tensor_names
+
+
+class TFModelImpl:
+    """
+    TFModelImpl is a wrapper on top of tensorflow which implements DLRModel API
+
+    Parameters
+    ----------
+    frozen_graph_file : str
+        Full path to frozen graph (.pb file)
+    dev_type : str
+        Optional. Device type ('cpu' or 'gpu')
+    dev_id : int
+        Optional. Device ID
+    """
+    def __init__(self, frozen_graph_file, dev_type=None, dev_id=None):
+        if not os.path.exists(frozen_graph_file):
+            raise ValueError("frozen_graph_file %s doesn't exist" % frozen_graph_file)
+        self.frozen_graph_file = frozen_graph_file
+
+        device = None
+        if dev_type is not None:
+            devices = ["cpu", "gpu"]
+            if dev_type not in devices:
+                raise ValueError("Invalid device type {}. Valid devices: {}".format(dev_type, devices))
+            dev_id = 0 if dev_id is None else dev_id
+            device = "/{}:{}".format(dev_type, dev_id)
+
+        self.graph = _load_frozen_graph(frozen_graph_file, device)
+        self.input_tensor_names, self.output_tensor_names = _get_input_and_output_names(self.graph)
+        self.input_values = {}
+
+    def _validate_input_name(self, name):
+        if name not in self.input_tensor_names:
+            raise ValueError(
+                "Invalid input tensor name '{}'. List of input tensor names: {}".format(name, self.input_tensor_names))
+
+    def _validate_input(self, input_values):
+        if isinstance(input_values, dict):
+            for k in input_values.keys():
+                if not isinstance(k, str):
+                    raise ValueError("input key must be string")
+                self._validate_input_name(k)
+        else:
+            raise ValueError("input_values must be of type dict")
+
+    def get_input_names(self):
+        """
+        Get all input names
+
+        Returns
+        -------
+        out : list of :py:class:`str`
+        """
+        return self.input_tensor_names
+
+    def get_output_names(self):
+        """
+        Get all output names
+
+        Returns
+        -------
+        out : list of :py:class:`str`
+        """
+        return self.output_tensor_names
+
+    def get_input(self, name, shape=None):
+        """
+        Get the current value of an input
+
+        Parameters
+        ----------
+        name : str
+            The name of an input
+        shape : np.array (optional)
+            If given, use as the shape of the returned array. Otherwise, the shape of
+            the returned array will be inferred from the last call to set_input().
+        """
+        self._validate_input_name(name)
+        if name not in self.input_values:
+            return None
+        out = self.input_values[name]
+        if shape is not None:
+            out = out.reshape(shape)
+        return out
+
+    def run(self, input_values):
+        """
+        Run inference with given input(s)
+
+        Parameters
+        ----------
+        input_values : a dictionary where keys are input
+            names (of type :py:class:`str`) and values are input tensors (of any type).
+            Multiple inputs are allowed.
+
+        Returns
+        -------
+        out : :py:class:`numpy.ndarray`
+            Prediction result. Multiple outputs are possible.
+        """
+        self._validate_input(input_values)
+        with tf.Session(graph=self.graph) as sess:
+            feed_dict = {}
+            for k, v in input_values.items():
+                tensor = self.graph.get_tensor_by_name(k)
+                feed_dict[tensor] = v
+            output_tensors = []
+            for k in self.output_tensor_names:
+                tensor = self.graph.get_tensor_by_name(k)
+                output_tensors.append(tensor)
+            self.input_values = input_values
+
+            out = sess.run(output_tensors, feed_dict=feed_dict)
+            return out
--- a/tests/python/unittest/test_tf_model.py
+++ b/tests/python/unittest/test_tf_model.py
@@ -1,0 +1,60 @@
+# coding: utf-8
+from dlr.tf_model import TFModelImpl
+import numpy as np
+
+FROZEN_GRAPH_PATH = "/tmp/test_graph.pb"
+
+
+def _generate_frozen_graph():
+    import tensorflow as tf
+    a = tf.placeholder(tf.float32, shape=[2, 2], name="input1")
+    b = tf.placeholder(tf.float32, shape=[2, 2], name="input2")
+    ab = tf.matmul(a, b)
+    mm = tf.matmul(a, ab)
+    tf.square(mm, name="output1")
+    mm_flat = tf.reshape(mm, shape=[-1])
+    tf.argmax(mm_flat, name="output2")
+
+    with tf.Session() as sess:
+        output_graph_def = tf.compat.v1.graph_util.convert_variables_to_constants(
+            sess,
+            tf.get_default_graph().as_graph_def(),
+            ["output1", "output2"]
+        )
+
+        with tf.gfile.GFile(FROZEN_GRAPH_PATH, "wb") as f:
+            f.write(output_graph_def.SerializeToString())
+    sess.close()
+
+
+def test_tf_model(dev_type=None, dev_id=None):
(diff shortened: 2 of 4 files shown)


python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
+# coding: utf-8
+import os
+import tensorflow as tf
+
+# A prefix that will be prepended to the names in graph_def
+PREFIX = "import"
```
yongwww: would be great to handle other prefix, as well as "no prefix". We met other prefix before actually

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class TFModelImpl:
+    """
+    TFModelImpl is a wrapper on top of tensorflow which implements DLRModel API
+
+    Parameters
+    ----------
+    frozen_graph_file : str
```
yongwww: frozen_graph_file -> model_path/model_dir? since probably we will support savedmodel or other types of format with this parameter

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+            if dev_type not in devices:
+                raise ValueError("Invalid device type {}. Valid devices: {}".format(dev_type, devices))
+            dev_id = 0 if dev_id is None else dev_id
+            device = "/{}:{}".format(dev_type, dev_id)
+
+        self.graph = self._load_frozen_graph(frozen_graph_file, device)
+        self.input_tensor_names, self.output_tensor_names = self._get_input_and_output_names(self.graph)
+        self.input_values = {}
+
+    @staticmethod
```
yongwww: don't need static here?

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+    def _load_frozen_graph(frozen_graph_file, device):
+        print("Loading frozen graph: {}".format(frozen_graph_file))
```
yongwww: better to use logging

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+        with tf.device(device):
+            with tf.gfile.GFile(frozen_graph_file, 'rb') as f:
+                graph_def = tf.GraphDef()
+                graph_def.ParseFromString(f.read())
+            with tf.Graph().as_default() as graph:
+                tf.import_graph_def(graph_def, name=PREFIX)
+            return graph
+
+    @staticmethod
```
yongwww: don't need static?

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+    def _get_input_and_output_names(graph):
+        input_tensor_names = []
+        output_tensor_names = set()
```
yongwww: output_tensor_names = []? since you cast set to list then

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+        op_prefix = PREFIX + "/"
+        for op in graph.get_operations():
+            if not op.name.startswith(op_prefix):
+                continue
+            if op.type == 'Placeholder' and op.inputs.__len__() == 0 and op.outputs.__len__() == 1:
```
yongwww: len(op.outputs) >= 1 ?

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+    def get_input(self, name, shape=None):
+        """
+        Get the current value of an input
+
+        Parameters
+        ----------
+        name : str
+            The name of an input
+        shape : np.array (optional)
+            If given, use as the shape of the returned array. Otherwise, the shape of
```
yongwww: add " Returns"

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        Get all output names
+
+        Returns
+        -------
+        out : list of :py:class:`str`
+        """
+        return self.output_tensor_names
+
+    def get_input(self, name, shape=None):
```
yongwww: confused about the function name, maybe get_intput_values?

tests/python/unittest/test_tf_model.py
```diff
@@ -0,0 +1,60 @@
...
+    with tf.Session() as sess:
+        output_graph_def = tf.compat.v1.graph_util.convert_variables_to_constants(
+            sess,
+            tf.get_default_graph().as_graph_def(),
+            ["output1", "output2"]
+        )
+
+        with tf.gfile.GFile(FROZEN_GRAPH_PATH, "wb") as f:
+            f.write(output_graph_def.SerializeToString())
+        print("%d ops in the final graph." % len(output_graph_def.node))
```
yongwww: log

tests/python/unittest/test_tf_model.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_tf_model(dev_type=None, dev_id=None):
```
yongwww: it would be good to add more test cases, to cover the pb file of a real model like mobilenet.

python/dlr/tf_model.py
apivovarov: Non empty Prefix is needed to correctly find input and output tensors from frozen graph. I saw that in some cases frozen graphs contain lots of operators without prefix which have tensors which might be wrongly identified as global input/output tensors.

python/dlr/tf_model.py
apivovarov: DLC compiler might support input different formats, e.g. saved model and frozen graph. After the "compilation" DLC will produce TF frozen graph as an output. So, `TFModelImpl` only needs to support frozen graph format.

python/dlr/tf_model.py
apivovarov: If class method does not use `self.` variables it's better to mark it as static (at least PyCharm suggests it this way).

python/dlr/tf_model.py
apivovarov: PyCharm said - method can be static

python/dlr/tf_model.py
apivovarov: `TFModelImpl` must implement all public method from DLRModel. DLRModel is out public  interface.

tests/python/unittest/test_tf_model.py
apivovarov: It will be done in the integration test. Out current approach with real model tests is to store model files on S3. Integration test downloads model file and runs the test.
This particular test is unit test. Unit tests should be small, fast and do not need internet connection.

python/dlr/tf_model.py
apivovarov: `api.py` should be fixes in the first place

tests/python/unittest/test_tf_model.py
```diff
@@ -0,0 +1,60 @@
...
+    assert inp_names == ['import/input1:0', 'import/input2:0']
+
+    out_names = m.get_output_names()
+    assert out_names == ['import/output1:0', 'import/output2:0']
+
+    inp1 = [[4., 1.], [3., 2.]]
+    inp2 = [[0., 1.], [1., 0.]]
+
+    res = m.run({'import/input1:0': inp1, 'import/input2:0': inp2})
+    print("Output: {}".format(res))
```
zhiics: remove?

python/dlr/tf_model.py
```diff
@@ -0,0 +1,153 @@
+# coding: utf-8
+import os
+import tensorflow as tf
+
+# A prefix that will be prepended to the names in graph_def
+PREFIX = "import"
+
+
+class TFModelImpl:
```
zhiics: Are you planning to implement the proposed RFC #46 later?

python/dlr/tf_model.py
apivovarov: I tried to use placeholder Tensor `input1:0` as an input for two other Tensors. When I listed graph_def operations I got `input1:0` operation and its outputs had only 1 item.
I think Placeholder operation type can only have one output tensor.
```
a = tf.placeholder(tf.float32, shape=[2, 2], name="input1")
print("--------------")
print(type(a))
b = tf.placeholder(tf.float32, shape=[2, 2], name="input2")
ab = tf.matmul(a, b)
mm = tf.matmul(a, ab)

for op in graph.get_operations():
    print(op, op.outputs)

name: "import/input1"
op: "Placeholder"
attr {
  key: "dtype"
  value {
    type: DT_FLOAT
  }
}
attr {
  key: "shape"
  value {
    shape {
      dim {
        size: 2
      }
      dim {
        size: 2
      }
    }
  }
}
 [<tf.Tensor 'import/input1:0' shape=(2, 2) dtype=float32>]
```

tests/python/unittest/test_tf_model.py
apivovarov: I think I can just remove this print

python/dlr/tf_model.py
apivovarov: fixed

tests/python/unittest/test_tf_model.py
apivovarov: removed

python/dlr/tf_model.py
apivovarov: `DLRModel` refactoring is a separate activity. It should not change `TFModelImpl`.
It will be another PR to refactor `DLRModel` as proposed in #46.  `DLRModel` will use `TFModelImpl` if model folder contains single `.pb` file

python/dlr/tf_model.py
yongwww: got it, i misunderstood outputs here

python/dlr/tf_model.py
```diff
@@ -0,0 +1,154 @@
+# coding: utf-8
+import logging
+import os
+import tensorflow as tf
+
+# A prefix that will be prepended to the names in graph_def
+PREFIX = "import"
+
+
+def _load_frozen_graph(frozen_graph_file, device):
```
yongwww: _load_frozen_graph -> load_frozen_graph, would be good to add description

python/dlr/tf_model.py
```diff
@@ -0,0 +1,154 @@
...
+    with tf.device(device):
+        with tf.gfile.GFile(frozen_graph_file, 'rb') as f:
+            graph_def = tf.GraphDef()
+            graph_def.ParseFromString(f.read())
+        with tf.Graph().as_default() as graph:
+            tf.import_graph_def(graph_def, name=PREFIX)
+        return graph
+
+
+def _get_input_and_output_names(graph):
```
yongwww: _get_input_and_output_names -> get_input_and_output_names, would be good to add description for this func

python/dlr/tf_model.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_input_names(self):
+        """
+        Get all input names
+
+        Returns
+        -------
+        out : list of :py:class:`str`
+        """
+        return self.input_tensor_names
```
kevinthesun: Better to set internal attribute self._input_tensor_names, and set this method as a property method.

python/dlr/tf_model.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_output_names(self):
+        """
+        Get all output names
+
+        Returns
+        -------
+        out : list of :py:class:`str`
+        """
+        return self.output_tensor_names
```
kevinthesun: Same here.

python/dlr/tf_model.py
```diff
@@ -0,0 +1,181 @@
...
+        Full path to frozen graph (.pb file)
+    dev_type : str
+        Optional. Device type ('cpu' or 'gpu')
+    dev_id : int
+        Optional. Device ID
+    """
+    def __init__(self, frozen_graph_file, dev_type=None, dev_id=None):
+        if not os.path.exists(frozen_graph_file):
+            raise ValueError("frozen_graph_file %s doesn't exist" % frozen_graph_file)
+        self.frozen_graph_file = frozen_graph_file
```
kevinthesun: Is this intended to be accessed externally? If not, we'd better set it to be internal.

python/dlr/tf_model.py
```diff
@@ -0,0 +1,181 @@
...
+
+        device = None
+        if dev_type is not None:
+            devices = ["cpu", "gpu"]
+            if dev_type not in devices:
+                raise ValueError("Invalid device type {}. Valid devices: {}".format(dev_type, devices))
+            dev_id = 0 if dev_id is None else dev_id
+            device = "/{}:{}".format(dev_type, dev_id)
+
+        self.graph = _load_frozen_graph(frozen_graph_file, device)
```
kevinthesun: Same here.

python/dlr/tf_model.py
apivovarov: added docs

python/dlr/tf_model.py
apivovarov: added docs

python/dlr/tf_model.py
apivovarov: renamed to `_input_tensor_names`

python/dlr/tf_model.py
apivovarov: `_output_tensor_names`

python/dlr/tf_model.py
apivovarov: renamed to `_frozen_graph_file`

python/dlr/tf_model.py
apivovarov: renamed to `self._graph`

python/dlr/tf_model.py
apivovarov: renamed `self.input_values` to `self._input_values`

apivovarov: As for public methods rename. The intention was to copy public api from `DLRModel` without any modifications. Duck typing. The idea is that end-users will use `DLRModel` as unified API. internally `DLRModel` will use different implementations depends on model file(s) format.
`DLRModel` refactoring and integration with `TFModelImpl` will be done in the next PR

kevinthesun: Per discussion, we can put refactor/redesign for DLRModel in another thread. For TFModelImpl, let's explicitly inherit it from DLRModel for now, so that it is clear these api should keep consistent.

python/dlr/tf_model.py
kevinthesun: Same above

python/dlr/tf_model.py
kevinthesun: API design will be completed for base interface in another PR. This is fine for now.

python/dlr/tf_model.py
kevinthesun: API design will be completed for base interface in another PR. This is fine for now.

yongwww:
Could you guys take another round of review? ping @wweic @zhiics @kevinthesun @hcho3

PR #49, opened 2019-04-29
Do not close TF session at the end of run method

apivovarov:
*Description of changes:*
When I run `mobilenet_v1_0.25_224_quant_frozen.pb` on ARMv8 RK3399 device I noticed that each run takes about 750 ms.
I refactored TFModelImpl code to not close TF session after each run(). After that change the second and consequent calls of run() method take just 130 ms.
```
m.run...
m.run done, duration 754 ms
m.run...
m.run done, duration 122 ms
m.run...
m.run done, duration 126 ms
m.run...
m.run done, duration 134 ms
m.run...
m.run done, duration 125 ms
```

I also implemented `__enter__` and `__exit__` methods in `TFModelImpl`.
It allows to use `TFModelImpl` with Context manager
```
with TFModelImpl(FROZEN_GRAPH_PATH, dev_type, dev_id) as m:
    for i in range(100):
        res = m.run(...)
```

### DIFF
--- a/tests/python/unittest/test_tf_model.py
+++ b/tests/python/unittest/test_tf_model.py
@@ -14,42 +14,38 @@ def _generate_frozen_graph():
     tf.square(mm, name="output1")
     mm_flat = tf.reshape(mm, shape=[-1])
     tf.argmax(mm_flat, name="output2")
-
     with tf.Session() as sess:
-        output_graph_def = tf.compat.v1.graph_util.convert_variables_to_constants(
+        output_graph_def = tf.graph_util.convert_variables_to_constants(
             sess,
             tf.get_default_graph().as_graph_def(),
             ["output1", "output2"]
         )
-
         with tf.gfile.GFile(FROZEN_GRAPH_PATH, "wb") as f:
             f.write(output_graph_def.SerializeToString())
-    sess.close()
 
 
 def test_tf_model(dev_type=None, dev_id=None):
     _generate_frozen_graph()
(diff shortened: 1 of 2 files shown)


apivovarov:
@yongwww @kevinthesun @wweic  Could you have a look?

tests/python/unittest/test_tf_model.py
```diff
@@ -14,42 +14,38 @@ def _generate_frozen_graph():
     tf.square(mm, name="output1")
     mm_flat = tf.reshape(mm, shape=[-1])
     tf.argmax(mm_flat, name="output2")
-
     with tf.Session() as sess:
-        output_graph_def = tf.compat.v1.graph_util.convert_variables_to_constants(
+        output_graph_def = tf.graph_util.convert_variables_to_constants(
```
apivovarov: nvidia only provides TF 1.10.0 for Jetson TX2
https://developer.download.nvidia.com/compute/redist/jp40/tensorflow-gpu/
```
Index of /compute/redist/jp40/tensorflow-gpu
..
tensorflow_gpu-1.10.0+nv18.9-cp27-cp27mu-linux_aarch64.whl 155MB 2018-09-14 14:23
tensorflow_gpu-1.10.0+nv18.9-cp36-cp36m-linux_aarch64.whl 155MB 2018-09-14 17:10
```
tf.compat.v1 was added after TF 1.10.0.
Because of that I switched to original deprecated API here.

apivovarov:
instead of running several requests in parallel user should batch them and run as one batch request. One N batch request will finish much faster than N single batch concurrent requests.

PR #51, opened 2019-05-02
Add multiple target support on Android

ziyu-guo:
Add x86 and aarch64 support on Android


CMakeLists.txt
```diff
@@ -13,7 +13,15 @@ if(ANDROID_BUILD)
...
     set(ANDROID_PLATFORM android-21)
     set(ANDROID_DEPRECATED_HEADERS ON)
-    include_directories(${ANDROID_SYSROOT}/usr/include/arm-linux-androideabi)
+    if (${ANDROID_ABI} STREQUAL "x86_64")
+        include_directories(${ANDROID_SYSROOT}/usr/include/x86_64-linux-android)
+    elseif (${ANDROID_ABI} STREQUAL "x86")
+        include_directories(${ANDROID_SYSROOT}/usr/include/i686-linux-android)
+    elseif (${ANDROID_ABI} STREQUAL "arm64-v8a")
+        include_directories(${ANDROID_SYSROOT}/usr/include/aarch64-linux-android)
+    else() # Default to armv7a which matches NDK toolchain.cmake behavior
```
zhiics: Will we only have the above cases? I think it is probably better to check armv7a and include the directories accordingly. The default case could just print error msg and exit.

ziyu-guo: Since we're using NDK's toolchain.cmake to control include/bin version, we're trying to be consistent with their behavior, and they do default to armv7a with 
`if(NOT ANDROID_ABI)
  set(ANDROID_ABI armeabi-v7a)
endif()`

PR #54, opened 2019-05-08
Unified dispatcher to 3rdparty packages

ziyu-guo:
*Issue #, if available:*

*Description of changes:*


python/dlr/api.py
```diff
@@ -342,3 +343,37 @@ def get_input(self, name, shape=None):
...
         out = out.reshape(shape)
         return out
+
+# Interface
+class IDLRModel:
+    __metaclass__=abc.ABCMeta
+
+    @abc.abstractmethod
+    def get_input_names(self):
+        return
```
wweic: throw https://docs.python.org/3/library/exceptions.html#NotImplementedError

tests/python/integration/load_and_run_tvm_model.py
```diff
@@ -27,8 +27,8 @@ def test_multi_input_multi_output():
     device = 'cpu'
     model = DLRModel(model_path, device)
 
-    assert model._get_output_size_dim(0) == (2, 1)
-    assert model._get_output_size_dim(1) == (3, 1)
+    #assert model._get_output_size_dim(0) == (2, 1)
```
wweic: why comment out this line?

python/dlr/api.py
zhiics: +1
do it for the other abstract methods that shouldn't be reached as well

tests/python/integration/load_and_run_tvm_model.py
ziyu-guo: _get_output_size_dim is TVM specific API and not available in new IDLRModel. Perhaps a better 
way is to simply delete these lines.

wweic:
I thought about it again. Could you add a unit test to test the support for tensorflow?

ziyu-guo:
> I thought about it again. Could you add a unit test to test the support for tensorflow? @ziyu-guo

@apivovarov Can you upload the the matching .pb to our test artifact S3 bucket? You can refer to test_utils.py for details.

tests/python/integration/load_and_run_tvm_model.py
zhiics: I think we should keep these assertions. We might want to use TVM object directly in this case.

python/dlr/api.py
```diff
@@ -71,7 +72,7 @@ def _check_call(ret):
     if ret != 0:
         raise DLRError(_LIB.DLRGetLastError().decode('ascii'))
 
-class DLRModel:
+class DLRModelImpl:
```
apivovarov: `DLRModel` dispatcher does not need `ctypes`, `numpy` and `_LIB = _load_lib()` in case we are working with `.pb` file.
I think `api.py` should contain minimum imports/code. It only needs the interface and the dispatcher.
I'd move `DLRModelImpl` and all its imports to `dlr_model.py` file (module).

apivovarov:
> I thought about it again. Could you add a unit test to test the support for tensorflow? @ziyu-guo

We do have `test_tf_model.py` module which contains unit tests which test that TFModelImpl can 
1. load frozen graph .pb file
2. correctly find input and output tensor names
3. run the model and evaluate output tensors
4. get input tensors values

apivovarov:
> > I thought about it again. Could you add a unit test to test the support for tensorflow? @ziyu-guo
> 
> @apivovarov Can you upload the the matching .pb to our test artifact S3 bucket? You can refer to test_utils.py for details.

That will be integration test, not unit test. We already have test_tf_model.py unit test case. Using unit test case is enough to test that DLRModel can load `.pb` file and `run()` the model correctly. It also tests that `get_input_names()` and `get_input()` work correctly for '.pb' file.

tests/python/integration/load_and_run_tvm_model.py
apivovarov: You can use `model._impl` instead of `model`
```
    assert model._impl._get_output_size_dim(0) == (2, 1)
    assert model._impl._get_output_size_dim(1) == (3, 1)
```

python/dlr/api.py
```diff
@@ -342,3 +343,41 @@ def get_input(self, name, shape=None):
...
+
+    @abc.abstractmethod
+    def run(self, input_data):
+        raise NotImplementedError
+
+# Wrapper class
+class DLRModel(IDLRModel):
+    def __init__(self, model_path, dev_type='cpu', dev_id=0):
+        #Determine if 3rdparty package needed
+        for (dirpath, dirnames, filenames) in os.walk(model_path):
```
apivovarov: This approach works only for folders. In case of TF model `model_path` will be (can be) .pb file (not folder)

python/dlr/api.py
```diff
@@ -342,3 +343,41 @@ def get_input(self, name, shape=None):
...
+            for filename in filenames:
+                if filename.endswith('.pb'): 
+                    from .tf_model import TFModelImpl
+                    self._impl = TFModelImpl(model_path, dev_type, dev_id)
```
apivovarov: if model_path is a folder then we should make sure the folder has only one `.pb` file.
BTW, I'd make support for both options:
1. for `.pb` file, e.g. `model_path=/tmp/mobilenet_v1/frozen_graph.pb`
2. for folder containing single pb file, e.g. `model_path=/tmp/mobilenet_v1`

python/dlr/api.py
ziyu-guo: Hi Alex, we can always put imports right before where they're used, so I think the real question is do we make dlr_model and tf_model equal implementations (your suggestion), or do we treat tf_model as an extension/bypass to dlr (current structure), since tf_model relies on external python pkgs. 

Thoughts?

wweic:
> We already have test_tf_model.py unit test case. Using unit test case is enough to test that DLRModel can load `.pb` file and `run()` the model correctly

@apivovarov `test_tf_model.py` only tests `TFModelImpl`, the test case you described is exactly what I'm asking for. :-)

python/dlr/api.py
apivovarov: I think that from architectural point of view DLRModelImpl is equal to TFModelImpl.

python/dlr/api.py
```diff
@@ -1,344 +1,50 @@
...
+class DLRModel(IDLRModel):
+    def __init__(self, model_path, dev_type='cpu', dev_id=0):
+        #Determine if 3rdparty package needed
+        if os.path.isfile(model_path):    
+            if model_path.endswith('.pb'):
+                from .tf_model import TFModelImpl
+                self._impl = TFModelImpl(model_path, dev_type, dev_id)
+        else:
+            for (dirpath, dirnames, filenames) in os.walk(model_path):
+                for filename in filenames:
```
apivovarov: Can we throw "Can not load the model" error if multiple pb files found

python/dlr/tf_model.py
```diff
@@ -132,7 +132,7 @@ def get_input_names(self):
         """
         return self.input_tensor_names
 
-    def get_output_names(self):
+    def _get_output_names(self):
```
apivovarov: Actually, it would be useful for the user to know what outputs are in the output.
e.g. 
```
output names: ['import/detection_boxes:0', 'import/detection_classes:0', 
    'import/detection_scores:0', 'import/num_detections:0']

# Having the info above helps users to understand what the output below is
[array([[[0.10138726, 0.39933354, 0.92921317, 0.9747917 ],
        [0.04678747, 0.01860756, 0.8673557 , 0.31017742],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ],
        [0.        , 0.        , 0.        , 0.        ]]], dtype=float32), array([[18., 18.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
         1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.]], dtype=float32), array([[0.95079756, 0.94477737, 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ]],
      dtype=float32), array([2.], dtype=float32)]
```
I'm planning to add this method to the interface. I think it is ok if we keep `get_output_names` as a public method in `TFModelImpl` and make it official in the interface later.

python/dlr/api.py
ziyu-guo: The current TVM logic in dlr.cc uses the first set of model artifacts found under directory (assumption is that the directory is either decompressed from tarball, or prepared in the same folder structure as the tarball). Best to keep the behaviors consistent.

python/dlr/api.py
```diff
@@ -17,6 +17,10 @@ def get_input_names(self):
     def get_input(self, name, shape=None):
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def get_output_names(self):
```
ziyu-guo: Per discussion with @apivovarov we're exposing this API and leave the TVM/Treelite impl as placeholder. We should open a discussion thread and possibly maintain a table showing which public DLR API corresponds to which API for each underlying runtime.

tests/python/unittest/test_tf_model.py
```diff
@@ -1,11 +1,10 @@
 # coding: utf-8
-from dlr.tf_model import TFModelImpl
+from __future__ import print_function
+from dlr import DLRModel
 import numpy as np
+import os
 
-FROZEN_GRAPH_PATH = "/tmp/test_graph.pb"
```
apivovarov: Replacing hardcoded `/tmp` with {current_dir} might be useful but I think it should be done in a separate PR.
I'd recommend to make changes in `tf_model.py` and  `test_tf_model.py` only if they are directly related to the goal of this PR, which is introduction of `IDLRModel` interface and `DLRModel` dispatcher

ziyu-guo: I made these changes just for that. We want to present TF and TVM/Treelite test files with uniform API and usage, as users will see them as our example.

ziyu-guo: Ok, I will only change needed API and add a main function.

PR #55, opened 2019-05-09
Add TFLiteModelImpl

apivovarov:
*Description of changes:*

`TFLiteModelImpl` is a wrapper on top of Tensorflow Lite which implements `IDLRModel` API

`test_tflite_model.py` contains unit tests. The unit test generates `/tmp/test_model.tflite` file and tests that DLRModel can:

1. load `/tmp/test_model.tflite` file
2. get input / output tensor names
3. run the model
4. get input values by name

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -135,7 +135,8 @@ def CloudInstallAndTest(cloudTarget) {
     sh """
     ls -lh *.whl
     pip3 install *.whl
-    pip3 install tensorflow
+    sudo pip3 install --upgrade --force-reinstall tensorflow
+    type toco_from_protos
     """
     echo "Running integration tests..."
     unstash name: 'srcs'
@@ -144,6 +145,7 @@ def CloudInstallAndTest(cloudTarget) {
     python3 tests/python/integration/load_and_run_treelite_model.py
     python3 -m pytest -v --fulltrace -s tests/python/unittest/test_get_set_input.py
     python3 -m pytest -v --fulltrace -s tests/python/unittest/test_tf_model.py
+    python3 -m pytest -v --fulltrace -s tests/python/unittest/test_tflite_model.py
     """
   }
 }
--- a/python/dlr/api.py
+++ b/python/dlr/api.py
@@ -3,8 +3,9 @@ from __future__ import absolute_import as _abs
 
 import abc
 import glob
+import logging
 import os
-import sys
+
 
 # Interface
 class IDLRModel:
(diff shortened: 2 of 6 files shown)


python/dlr/api.py
```diff
@@ -26,27 +26,41 @@ def get_output_names(self):
...
+        # Tensorflow
+        tf_model_path = _find_model_file(model_path, '.pb')
         if tf_model_path is not None:
             from .tf_model import TFModelImpl
             self._impl = TFModelImpl(tf_model_path, dev_type, dev_id)
-        else:
-            from .dlr_model import DLRModelImpl
-            self._impl = DLRModelImpl(model_path, dev_type, dev_id) 
+            return
+        # TFLite
```
yongwww: if we find both pb and tflite, warn user which one is used or just raise error?

python/dlr/tflite_model.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Parameters
+    ----------
+    intrp : :py:class:`str`tf.lite.Interpreter`
+        TFLite Interpreter
+
+    Returns
+    -------
+    input_tensor_map : Input tensor Map which maps tensor_name to tensor_id
+    output_tensor_list  : Output tensor list. List element is pair (name, id)
```
yongwww:
```suggestion
    output_tensor_list : Output tensor list. List element is pair (name, id)
```

python/dlr/api.py
apivovarov: fixed. will throw the Exception if both found

python/dlr/tflite_model.py
apivovarov: fixed

Jenkinsfile
```diff
@@ -135,7 +135,8 @@ def CloudInstallAndTest(cloudTarget) {
     sh """
     ls -lh *.whl
     pip3 install *.whl
-    pip3 install tensorflow
+    sudo pip3 install tensorflow
```
apivovarov: need sudo here because it installs `toco_from_protos` to `/usr/local/bin/`

zhiics: Trying to understand, why `toco_from_protos ` has to be installed under `/usr/local/bin`? Can we use `pip3 install --user tensorflow`?

python/dlr/api.py
zhiics: I am okay with raising an error, but I think warning (then using TF Lite by default) might be a better option.

python/dlr/api.py
```diff
@@ -26,27 +26,44 @@ def get_output_names(self):
...
+        if tflite_model_path is not None:
+            from .tflite_model import TFLiteModelImpl
+            self._impl = TFLiteModelImpl(tflite_model_path, dev_type, dev_id)
+            return
```
zhiics: Is it possible that both of them are `None`? If so, we may need to throw an error.

python/dlr/tflite_model.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Parameters
+    ----------
+    intrp : :py:class:`str`tf.lite.Interpreter`
+        TFLite Interpreter
+
+    Returns
+    -------
+    input_tensor_map : Input tensor Map which maps tensor_name to tensor_id
+    output_tensor_list : Output tensor list. List element is pair (name, id)
```
zhiics: Why one is a map, but the other is list?

python/dlr/tflite_model.py
```diff
@@ -0,0 +1,133 @@
...
+
+class TFLiteModelImpl(IDLRModel):
+    """
+    TFLiteModelImpl is a wrapper on top of Tensorflow Lite which implements IDLRModel API
+
+    Parameters
+    ----------
+    tflite_file : :py:class:`str`
+        Full path to TFLite file (.tflite file)
+    dev_type : :py:class:`str`
```
zhiics: It looks that `dev_type` and `dev_id` are TVM specific. We probably don't need to put them in the base class. This reduces confusion for users.

python/dlr/tflite_model.py
```diff
@@ -0,0 +1,133 @@
...
+    def get_input_names(self):
+        """
+        Get all input names
+
+        Returns
+        -------
+        out : list of :py:class:`str`
+        """
+        names = list(self._input_tensor_map.keys())
+        names.sort()
```
zhiics: What's the original oder? Do we need to sort them, or we can make sorting optional to the users (i.e. `get_input_name(self, sorted=False)`?

python/dlr/api.py
apivovarov: We agreed to raise the error here

Jenkinsfile
apivovarov: Since it is in Jenkinsfile and used only by CI we agreed to keep sudo

python/dlr/api.py
apivovarov: if both TF and TFLite files are not found then the dispatcher will use DLRModelImpl by default

python/dlr/tflite_model.py
apivovarov: changed both input_tensor_map and output_tensor_list to OrderedDict. It allows to do lookup by tensor name and it keep tensor names in the original order returned TFLite Interpreter

python/dlr/tflite_model.py
apivovarov: Removed `dev_type` and `dev_id` from `TFLiteImpl.init`.
If user explicitly specifies `dev_type` or/and `dev_id` for `.tflite` file then the dispatcher will log Warning and ignore `dev_type` and `dev_id` parameters

python/dlr/tflite_model.py
apivovarov: removed `names.sort()` and use OrderedDict for `_input_tensor_map` instead

tests/python/unittest/test_tf_model.py
```diff
@@ -59,6 +59,7 @@ def test_tf_model_on_cpu_0():
 def test_tf_model_on_gpu_0():
     test_tf_model("gpu", 0)
 
+
 if __name__ == '__main__':
     test_tf_model()
     print('All tests passed!')
```
yongwww: delete?

yongwww:
```suggestion

```

PR #57, opened 2019-05-14
edit docker entry bash file to set threads accordingly

zxy844288792:
*Description of changes:*
set OMP_NUM_THREADS to num of cores

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -127,6 +127,7 @@ def CloudInstallAndTest(cloudTarget) {
     echo "Installing DLR package for ${cloudTarget} target"
     withAWS(credentials:'Neo-AI-CI-Fleet') {
       files = s3FindFiles(bucket: 'neo-ai-dlr-jenkins-artifacts', glob: "${env.JOB_NAME}/${env.BUILD_ID}/artifacts/dlr-*-manylinux1_x86_64.whl")
+      echo files.size()
       assert files.size() == 1
       files.each {
         s3Download file: it.name, bucket: 'neo-ai-dlr-jenkins-artifacts', path: it.path, force: true
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -129,8 +129,16 @@ def CloudInstallAndTest(cloudTarget) {
     sh """
     ls -lh python/dist/*.whl
     pip3 install python/dist/*.whl
-    pip3 install tensorflow
     """
+    if (cloudTarget == "p2" || cloudTarget == "p3") {
+      sh """
+      pip3 install tensorflow_gpu
+      """
+    } else {
+      sh """
+      pip3 install tensorflow
+      """
+    }
     echo "Running integration tests..."
     unstash name: 'srcs'
     sh """
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -98,9 +98,7 @@ def AMD64BuildCPU() {
     tests/ci_build/ci_build.sh ${dockerTarget} ${dockerArgs} tests/ci_build/build_via_cmake.sh
     tests/ci_build/ci_build.sh ${dockerTarget} ${dockerArgs} tests/ci_build/create_wheel.sh manylinux1_x86_64
     """
-    withAWS(credentials:'Neo-AI-CI-Fleet') {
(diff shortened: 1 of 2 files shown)


container/dockerd-entrypoint.sh
```diff
@@ -2,14 +2,26 @@
...
+    shift 1
     cp -v /opt/ml/model/* /home/model-server/model
     cp -v /home/model-server/neo_template.py /home/model-server/model
     model-archiver --handler neo_template:predict --model-name neomodel --model-path /home/model-server/model -f --export-path /home/model-server
     mv /home/model-server/neomodel.mar /home/model-server/model
 
-    shift 1
-    mxnet-model-server --start --mms-config config.properties
+    if [ ! -z "${MMS_NUM_WORKER}" ]
+    then
```
apivovarov: line 8 has `if` and `then` on the same line. Can you do the same on line 15?

zxy844288792: sure

Jenkinsfile
```diff
@@ -125,16 +121,14 @@ def CloudInstallAndTest(cloudTarget) {
...
+      unstash 'dlr_gpu_whl'
+    } else {
+      unstash 'dlr_cpu_whl'
     }
     sh """
-    ls -lh *.whl
-    pip3 install *.whl
+    ls -lh python/dist/*.whl
+    pip3 install python/dist/*.whl
     pip3 install tensorflow
```
yongwww: the command to install tf for gpu is `pip3 install tensorflow_gpu`

container/dockerd-entrypoint.sh
```diff
@@ -2,14 +2,25 @@
 set -e
 set -x
 
+export IFS=' '
+MMS_NUM_WORKER=1
```
yongwww: is MMS_NUM_WORKER configurable? one worker for one instance as default?

Jenkinsfile
```diff
@@ -125,18 +121,24 @@ def CloudInstallAndTest(cloudTarget) {
   def nodeReq = "ubuntu && amd64 && ${cloudTarget}"
   node(nodeReq) {
     echo "Installing DLR package for ${cloudTarget} target"
-    withAWS(credentials:'Neo-AI-CI-Fleet') {
-      files = s3FindFiles(bucket: 'neo-ai-dlr-jenkins-artifacts', glob: "${env.JOB_NAME}/${env.BUILD_ID}/artifacts/dlr-*-manylinux1_x86_64.whl")
-      assert files.size() == 1
-      files.each {
-        s3Download file: it.name, bucket: 'neo-ai-dlr-jenkins-artifacts', path: it.path, force: true
-      }
+    if (cloudTarget == "p2" || cloudTarget == "p3") {
```
apivovarov:
```
>>> "p2" in ("p2", "p3")
True
```

hcho3: Is this Groovy?

Jenkinsfile
```diff
@@ -125,18 +121,24 @@ def CloudInstallAndTest(cloudTarget) {
...
+    ls -lh python/dist/*.whl
+    pip3 install python/dist/*.whl
     """
+    if (cloudTarget == "p2" || cloudTarget == "p3") {
+      sh """
+      pip3 install tensorflow_gpu
+      """
+    } else {
+      sh """
+      pip3 install tensorflow
```
apivovarov: Just merged PR #55 where I changed the command to
```sudo pip3 install --upgrade --force-reinstall tensorflow```

yongwww:
@apivovarov  for tf installation, I remember you have specific tf binary/wheel for some instances, pls suggest or submit a follow-up pr after this.

container/dockerd-entrypoint.sh
zxy844288792: Yes it is, but right now we are keeping it for 1 for best latency. (I replied in the morning but I could not find my reply)

Jenkinsfile
zxy844288792: change to tensorflow_gpu for p3 and p2

Jenkinsfile
zxy844288792: I have just merged the latest commit

hcho3:
@yongwww Eventually, we should do all our tests inside a Docker container, so that we have reproducible installation and build. In particular, we can store the exact version of Tensorflow inside the Dockerfile.

PR #66, opened 2019-06-13
support heterogeneous models

ziyu-guo:
Support heterogeneous models without changing python or c API.

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -5,7 +5,7 @@
 #include <tvm/runtime/module.h>
 #include <treelite/predictor.h>
 #include <treelite/c_api_runtime.h>
-#include "../3rdparty/treelite/dmlc-core/src/io/filesys.h"
+#include "../3rdparty/treelite/dmlc-core/include/dmlc/filesystem.h"
 //#include <graph/graph_runtime.cc>
 //#include <graph/debug/graph_runtime_debug.cc>
 #include <runtime_base.h>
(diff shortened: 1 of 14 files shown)


include/dlr.h
```diff
@@ -15,6 +15,19 @@
 #include <sys/types.h>
 
 #define LINE_SIZE 256
+/*!
+ * \brief The DLContext device types supported by DLR.
+ */
+typedef enum {
```
zhiics: We don't have to redefine this. They are defined here:

https://github.com/dmlc/dlpack/blob/bee4d1dd8dc1ee4a1fd8fa6a96476c2f8b7492a3/include/dlpack/dlpack.h

ziyu-guo: This is to track, out of DLDeviceType, how many device types DLR currently run on. These supported devices are then all sent to underlying TVM runtime during construction. I put them here so if DLR supports new device, we don't need to change following lines in [dlr.cc](https://github.com/ziyu-guo/neo-ai-dlr/blob/e8b1aeb6fbc6ddc5636d62327830e7a1c29ff411/src/dlr.cc#L576-L578)

ziyu-guo:
We should probably port RuntimeEnabled() to DLR.

zhiics:
tvm python side has this API: https://github.com/dmlc/tvm/blob/master/python/tvm/module.py#L263

It looks me that we also need `time_evalutaor`

Issue #61, opened 2019-05-23
Add C++ test/example using C API


ziyu-guo:
Currently DLR only has tests/examples in python API. A C-API-based test needed for users to reference.

PR #67, opened 2019-06-13
add a self-contained cpp test

ziyu-guo:
To cover #61


tests/cpp/test_4in2out.cc
```diff
@@ -0,0 +1,106 @@
...
+std::vector<std::vector<float>> RunInference(DLRModelHandle model);
+
+// Based on hcho3's test harness, this executable 
+// load and test a simple TVM model that sums 2 pairs of tensors 
+// and outputs 2 result tensors.
+//    data1 data2     data3 data4
+//       \  /             \  /
+//      output1          output2
+
+int main(int argc, char** argv) {
```
zhiics: I thin this is not a good way for unit test. We shouldn't expect developers to explicitly provide anything. Let's use gtest directly for different testing cases.

You can find some examples from here:
https://github.com/dmlc/tvm/tree/master/tests/cpp

tests/cpp/gen_4in2out.py
```diff
@@ -0,0 +1,46 @@
...
+import sys
+
+import nnvm.symbol as symbol
+import nnvm.graph as graph
+import nnvm.compiler
+from nnvm.frontend import from_mxnet
+
+import tvm
+
+data1 = symbol.Variable(name="data1")
```
zhiics: Why do we need this file? It looks to me that they just generate the artifact. We probably should generate them only when we are doing the tests.

CMakeLists.txt
```diff
@@ -203,22 +207,25 @@ add_custom_target(combined_lib
 
 # Tests
 set(TEST_EXECS "")
-file(GLOB TEST_SRCS test/cpp/*.cc)
+file(GLOB TEST_SRCS tests/cpp/*.cc)
 find_library(GTEST_LIB NAMES libgtest.a gtest)
 
-if(GTEST_LIB)
```
zhiics: Why do we remove gtest? I think we need to keep them. And I think we need to tests for DLR using gtests.

zhiics:
Does #75 depend on this?

ziyu-guo:
> Does #75 depend on this?

No. #75 is a follow-up after addressing missing test_assign_op artifacts on arm devices.

CMakeLists.txt
ziyu-guo: This c test is supposed to be run on both linux and android, using gtest adds an additional dependency.

PR #70, opened 2019-06-19
Update api.py _find_model_file

zxy844288792:
*Issue #, if available:*
If we only specify the folder path instead of the pb file path, it will not find the file.

*Description of changes:*
Make sure it can fild the pb or tflite file if we only specify the folder path

### DIFF
--- a/python/dlr/api.py
+++ b/python/dlr/api.py
@@ -31,7 +31,7 @@ class IDLRModel:
 def _find_model_file(model_path, ext):
     if os.path.isfile(model_path) and model_path.endswith(ext):
         return model_path
-    model_files = glob.glob(os.path.abspath(os.path.join(model_path, ext)))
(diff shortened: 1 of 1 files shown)


zxy844288792:
> Need to handle [`if os.path.isfile(model_path) and model_path.endswith(ext):`](https://github.com/neo-ai/neo-ai-dlr/blob/c1ce310454ee7cb47186368232004723fe639924/python/dlr/api.py#L32)?

zxy844288792:
> Need to handle [`if os.path.isfile(model_path) and model_path.endswith(ext):`](https://github.com/neo-ai/neo-ai-dlr/blob/c1ce310454ee7cb47186368232004723fe639924/python/dlr/api.py#L32)?
Yes, you are right, going to fix it right now.
Thanks!

zxy844288792:
update to make sure it will acecpt both pb file path or folder path as input.

PR #76, opened 2019-07-18
Support multi-line request in XGBoost container

hcho3:
*Description of changes:* Current XGBoost container returns a 503 if the user submits a request with multiple data points. Fix the MMS-interfacing code to support this use case.

*How to test*: In Python shell, run:
```python
import requests

# Open a LIBSVM file with multiple data points (rows)
with open('tests/python/integration/xgboost/letor.libsvm', 'r') as f:
    data = f.read()

# Send request to a locally running Docker container
r = requests.post('http://localhost:8888/invocations', data=data,
                  headers={'Content-type': 'text/libsvm'})
```

### DIFF
--- a/container/neo_template_xgboost.py
+++ b/container/neo_template_xgboost.py
@@ -3,6 +3,7 @@ import pickle as pkl
 import csv
 import numpy as np
 from scipy.sparse import csr_matrix
+import io
 import os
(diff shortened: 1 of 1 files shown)


hcho3:
@zxy844288792 Can you review?

PR #79, opened 2019-07-27
Add WITH_TENSORFLOW_LITE_LIB cmake parameter

apivovarov:
Parameter `WITH_TENSORFLOW_LITE_LIB` allows to include pre-built `libtensorflow-lite.a` library to DLR shared library `libdlr.so`.

Example:
```
mkdir build
cd build

cmake .. -DANDROID_BUILD=ON \
-DNDK_ROOT=/opt/android-ndk-r18b/ \
-DCMAKE_TOOLCHAIN_FILE=/opt/android-ndk-r18b/build/cmake/android.toolchain.cmake \
-DANDROID_ABI=x86 \
-DCMAKE_CXX_FLAGS="-march=silvermont -mtune=silvermont -O3" \
-DWITH_TENSORFLOW_LITE_LIB="/opt/tensorflow/tensorflow/lite/tools/make/gen/i686-linux-android_silvermont/lib/libtensorflow-lite.a"

make -j8 VERBOSE=1

ls -la ../lib
-rwxrwxr-x  1 dlc dlc 18,702,416 Jul 26 20:00 libdlr.so
-rw-rw-r--  1 dlc dlc  3,527,988 Jul 26 20:00 libdlr_static.a
```


CMakeLists.txt
```diff
@@ -170,6 +170,10 @@ if(USE_TENSORRT)
     add_definitions(-DTVM_GRAPH_RUNTIME_TENSORRT)
     set(USE_TENSORRT OFF)
 endif()
+if(WITH_TENSORFLOW_LITE_LIB)
+    message("Adding TENSORFLOW_LITE_LIB to DLR shared library")
+    list(APPEND TVM_RUNTIME_LINKER_LIBS -llog -Wl,--whole-archive ${WITH_TENSORFLOW_LITE_LIB} -Wl,--no-whole-archive)
```
zhiics: is the second `-Wl` superfluous?

hcho3: @zhiics No, it is not, because we want apply `--whole-archive` option only to the TFLite lib and not to other libs.

PR #80, opened 2019-07-28
Fix configuration for MMS to optimize XGBoost inference performance

hcho3:
For XGBoost models, raw inference time is minuscule compared to serving overheads. So to best utilize multi-core CPUs, it's actually better to allocate many worker processes, and give each process one thread. This is quite different from deep learning workloads, where we allocate 1 single worker and give it many threads.

* Make Treelite runtime aware of environment variable `OMP_NUM_THREADS`.
* Use different process X thread configurations for XGBoost and Deep Learning

Benchmark results

* Latency below is measured for each batch, where each batch contains 100 rows
* Model used : learning-to-rank with YAHOO dataset, 1600 trees, 6 max depth
* Latency was measured by running inference containers in C5.9xlarge machine; the client (Apache Bench) was co-located with the inference container, so as to eliminate network latency.

**Median (P50) Latency**

|# concurrent requests| Neo (current)| Neo (PROPOSED FIX)|
|---|---|---|
|1|63|55|
|2|124|56|
|5|311|57|
|10|622|58|
		
**Throughput (QPS)**

|# concurrent requests| Neo (current)| Neo (PROPOSED FIX)|
|---|---|---|
|1|15.81|18.16|
|2|16.09|35.45|
|5|16.08|86.37|
|10|16.07|168.22|

It is quite clear that for XGBoost, serving takes vast majority of time (> 90%), not raw inference. So it makes sense to allocate lots of worker processes, so that we can handle high number of concurrent requests.

### DIFF
--- a/container/mms_config_image_classification.sh
+++ b/container/mms_config_image_classification.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+MMS_NUM_WORKER=1
+MMS_NUM_THREAD_PER_WORKER=`nproc`
--- a/container/Dockerfile.cpu
+++ b/container/Dockerfile.cpu
@@ -46,6 +46,7 @@ COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
 COPY config.properties /home/model-server/config.properties
 
 COPY neo_template_$APP.py /home/model-server/neo_template.py
+COPY mms_config_$APP.sh /home/model-server/mms_config.sh
 
 RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh \
     && chown -R model-server /home/model-server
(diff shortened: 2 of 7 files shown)


kevinthesun:
This should also apply for resnet case?

container/mms_config_image_classification.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+MMS_NUM_WORKER=1
+MMS_NUM_THREAD_PER_WORKER=`nproc`
```
kevinthesun: What is the number of workers setting before this change?

hcho3: https://github.com/neo-ai/neo-ai-dlr/blob/7d56499a210dcd0bea077a52914576d6fb721e89/container/dockerd-entrypoint.sh#L5-L6

hcho3:
> This should also apply for resnet case?

@kevinthesun No, the fix only applies to XGBoost, since the inference workload is so lightweight compared to model serving overhead:

Batch Size| Raw inference (ms)| End-to-end latency (ms)
---|---|---
1 |0.362238|3.158981
10 |0.355123 |6.127867
100|1.204103| 67.687982

PR #81, opened 2019-08-21
Move internal objects from dlr.h to dlr_internal.h

apivovarov:
This PR move internal objects from `dlr.h` to `dlr_internal.h`.
`dlr_internal.h` contains internal C++ objects needed internally for `dlr.cc`

Having liteweight `dlr.h` allows:
1. Use `dlr.h` file inside pure C programs
2. Use only one include folder `-Ineo-ai-dlr/include` during program compilation (before we had to include 5 additional folders from `3rdparty`)
3. `dlr.h` can be used by both C and C++ programs (`extern "C"` prefix is added to DLR functions for cplusplus case automatically)

C program example `run-resnet.c`:
```
#include <sys/types.h>
#include <string.h>
#include <stdio.h>
#include <sys/time.h>
#include <stdlib.h>
#include "dlr.h"
...
```

To compile C programm:
```
gcc -O2 -std=c11 -I/opt/neo-ai-dlr/include \
run-resnet.c -o run-resnet \
-L/opt/neo-ai-dlr/lib -ldlr -lpthread
```

To compile C++ programm:
```
g++ -O2 -std=c++11 -I/opt/neo-ai-dlr/include \
run-resnet.cc -o run-resnet \
-L/opt/neo-ai-dlr/lib -ldlr -lpthread
```

Full `run-resnet.c` - https://gist.github.com/apivovarov/146a7d32bc590eaca45024a1b3fa26ab
`run-resnet.c` extension can be changed to `.cc` and successfully compiled with `g++`

### DIFF
--- a/include/dlr_internal.h
+++ b/include/dlr_internal.h
@@ -1,0 +1,121 @@
+#ifndef DLR_INTERNAL_H_
+#define DLR_INTERNAL_H_
+
+#include <graph/graph_runtime.h>
+#include <tvm/runtime/module.h>
+#include <treelite/predictor.h>
+#include <treelite/c_api_runtime.h>
+#include "../3rdparty/treelite/dmlc-core/src/io/filesys.h"
+#include <runtime_base.h>
+#include <dlpack/dlpack.h>
+#include <string>
+#include <vector>
+#include <sys/types.h>
+#include "dlr.h"
+
+#define LINE_SIZE 256
+
+
+typedef struct {
+  std::string model_lib;
+  std::string params;
+  std::string model_json;
+  std::string ver_json;
+} ModelPath;
+
+void listdir(const std::string& dirname, std::vector<std::string> &paths);
+
+inline bool endsWith(const std::string &mainStr, const std::string &toMatch)
+{
+	if(mainStr.size() >= toMatch.size() &&
+	  mainStr.compare(mainStr.size() - toMatch.size(), toMatch.size(), toMatch) == 0)
+		return true;
+	else
+		return false;
+}
+
+enum class DLRBackend {
+  kTVM,
+  kTREELITE
+};
+/*! /brief Get the paths of the TVM model files.
+ */
+ModelPath get_tvm_paths(const std::string &tar_path);
+
+/*! /brief Get the paths of the Treelite model files.
+ */
+ModelPath get_treelite_paths(const std::string& dirname);
+
+/*! /brief Get the backend based on the contents of the model folder.
+ */
+DLRBackend get_backend(const std::string &dirname);
+
+namespace dlr {
+
+struct TreeliteInput {
+  std::vector<float> data;
+  std::vector<uint32_t> col_ind;
+  std::vector<size_t> row_ptr;
+  size_t num_row;
+  size_t num_col;
+  CSRBatchHandle handle;
+};
+
+class DLRModel {
+ private:
+  std::string version_;
+  DLRBackend backend_;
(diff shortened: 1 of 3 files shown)


include/dlr_internal.h
```diff
@@ -0,0 +1,123 @@
+#ifndef DLR_INTERNAL_H_
+#define DLR_INTERNAL_H_
+
+#include <graph/graph_runtime.h>
+#include <tvm/runtime/module.h>
+#include <treelite/predictor.h>
+#include <treelite/c_api_runtime.h>
+#include "../3rdparty/treelite/dmlc-core/src/io/filesys.h"
+//#include <graph/graph_runtime.cc>
```
zhiics: Let's remove these two commented lines.

include/dlr_internal.h
```diff
@@ -0,0 +1,123 @@
...
+		return false;
+}
+
+enum class DLRBackend {
+  kTVM,
+  kTREELITE
+};
+/*! /brief Get the paths of the TVM model files.
+ */
+ModelPath get_tvm_paths(const std::string &tar_path);
```
zhiics: Let's use the left pointer alignment style (`const std::string& tar_path`) to keep the consistency.

include/dlr_internal.h
```diff
@@ -0,0 +1,123 @@
...
+
+/*! /brief Get the paths of the Treelite model files.
+ */
+ModelPath get_treelite_paths(const std::string& dirname);
+
+/*! /brief Get the backend based on the contents of the model folder.
+ */
+DLRBackend get_backend(const std::string &dirname);
```
zhiics: Let's use the left pointer alignment style (`const std::string& dirname`) to keep the consistency.

include/dlr_internal.h
```diff
@@ -0,0 +1,123 @@
...
+  /*! /brief Load model files from given folder path.
+   */
+  explicit DLRModel(const std::string& model_path,
+                    const DLContext& ctx);
+
+  /*! /brief Get the output of the given input x.
+   */
+  void Run();
+
+  void SetupTVMModule(const std::string& model_path);
```
zhiics: I know these functions were not documented before. Do you mind documenting them?

include/dlr_internal.h
apivovarov: Removed

include/dlr_internal.h
apivovarov: There are 9 places in DLR `.cc` and `.h` files where we use `string &<var>` and 100+ places in `3rdparty` where they use `string &<var>`.
Lets discuss it.
Common recommendation is to put `*` in pointer declarations to the variable
https://stackoverflow.com/questions/180401/placement-of-the-asterisk-in-pointer-declarations
```
int *pointer1, *pointer2; // Fully consistent, two pointers
int* pointer1, pointer2;  // Inconsistent (because only the first one is a pointer, the second one is an int variable), unexpected, and thus prone to errors
```

include/dlr_internal.h
apivovarov: This is internal function used in `dlr.cc` file only.
We might even remove `dlr_internal.h` file in future since it contains only internal C++ declarations used only in `dlr.cc`.

include/dlr_internal.h
ziyu-guo: Agreed, I think we should use left alignment and avoid all multi-decl in the same line. 

https://dev.to/codemouse92/cc-pointer-alignment-style-a-justification-13po

PR #92, opened 2019-09-20
add c-api demo

ziyu-guo:
*Issue #, if available: #61 

*Description of changes:*

Added cmake target demo, which builds 2 linux/android executables: model_peeker and run_resnet. 

RunInference implementation adapted from hcho3's DLR_Echodot_Validation package. 

Usage: 
./model_peeker  [model dir] <device_type>

./run_resnet [model dir] [ndarray file] <device> <input name>

Tested on P3, Jetson and QC devices


demo/cpp/model_peeker.cc
```diff
@@ -0,0 +1,99 @@
...
+
+  input_names.resize(num_inputs);
+  std::cout << "input_names: ";
+  for (int i = 0; i < num_inputs; i++) {
+    GetDLRInputName(&model, i, &input_names[i]);
+    std::cout << input_names[i] << ", ";
+  }
+  std::cout << std::endl;
+  
+  // weight_names.resize(num_weights);
```
zhiics: Remove?

demo/cpp/model_peeker.cc
```diff
@@ -0,0 +1,99 @@
...
+    }
+    std::cout << "]" << std::endl;
+  }
+}
+
+int main(int argc, char** argv) {
+  int device_type = 1;
+  std::string input_name = "data";
+  if (argc < 2) {
+    std::cerr << "Usage: " << argv[0] << " <model dir> [device_type]" << std::endl;
```
zhiics: LOG(FATAL)?

demo/cpp/model_peeker.cc
```diff
@@ -0,0 +1,99 @@
...
+  if (argc >= 3) {
+    std::string argv2(argv[2]);
+    if (argv2 == "cpu"){
+      device_type = 1;
+    } else if (argv2 == "gpu") {
+      device_type = 2;
+    } else if (argv2 == "opencl") {
+      device_type = 4;
+    } else {
+      std::cerr << "Unsupported device type!" << std::endl;
```
zhiics: LOG(FATAL)?

demo/cpp/model_peeker.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <numeric>
+#include <algorithm>
+#include <functional>
+#include <vector>
+#include <limits>
+#include <stdexcept>
+#include <dlr.h>
+#include "dmlc/logging.h"
+
+/*! \brief Prints model metadata using DLR C-API.
```
zhiics: I don't have strong preference. Do you want to add it to the test pipeline (using gtest or could a script to build and run it somehow). Otherwise, if ppl change DLR but forgot to update this demo, it may just fail later when other ppl use it.

demo/cpp/npy.hpp
```diff
@@ -0,0 +1,517 @@
...
+
+const char magic_string[] = "\x93NUMPY";
+const size_t magic_string_length = 6;
+
+const char little_endian_char = '<';
+const char big_endian_char = '>';
+const char no_endian_char = '|';
+
+constexpr char host_endian_char = ( big_endian ? 
+    big_endian_char :
```
zhiics: align?

demo/cpp/npy.hpp
```diff
@@ -0,0 +1,517 @@
...
+
+  if (0 != std::memcmp(buf, magic_string, magic_string_length))
+    throw std::runtime_error("this file does not have a valid npy format.");
+
+  v_major = buf[magic_string_length];
+  v_minor = buf[magic_string_length+1];
+}
+
+// typestring magic
+struct Typestring {
```
zhiics: `struct` is usually used as a data container.  We probably want to use `class` in this case or make the members public.

demo/cpp/npy.hpp
```diff
@@ -0,0 +1,517 @@
...
+  std::regex_match(typestring, sm, re );
+
+  if ( sm.size() != 4 ) {
+    throw std::runtime_error("invalid typestring");
+  }
+}
+
+namespace pyparse {
+
+/**
```
zhiics: Could you please make the comments of functions as following?

```c++
/*!
 * \brief
 */
```
Or other format like `//`, but let's keep them consistent.

demo/cpp/npy.hpp
```diff
@@ -0,0 +1,517 @@
...
+   Parses the string representation of a Python dict
+
+   The keys need to be known and may not appear anywhere else in the data.
+ */
+inline std::unordered_map<std::string, std::string> parse_dict(std::string in, std::vector<std::string>& keys) {
+
+  std::unordered_map<std::string, std::string> map;
+
+  if (keys.size() == 0)
```
zhiics: keys.empty()?

demo/cpp/npy.hpp
```diff
@@ -0,0 +1,517 @@
...
+inline ndarray_len_t comp_size(const std::vector<ndarray_len_t>& shape) {
+    ndarray_len_t size = 1;
+    for (ndarray_len_t i : shape )
+      size *= i;
+
+    return size;
+}
+
+template<typename Scalar>
+inline void SaveArrayAsNumpy( const std::string& filename, bool fortran_order, unsigned int n_dims, const unsigned long shape[], const std::vector<Scalar>& data)
```
zhiics: format this line?

demo/cpp/model_peeker.cc
ziyu-guo: The demo is using only core APIs that are intended to be alway backward-compatible. Will add C-API gtests in following PRs.

demo/cpp/npy.hpp
ziyu-guo: npy.hpp is 3rd party utility code that we don't want to change.

demo/cpp/model_peeker.cc
ziyu-guo: Do you mean the weight names prints? I left them commented out because while they're using valid DLR C-API, for most models there're just too many weight nodes that the output info makes less sense. But user can still add them back easily.

PR #93, opened 2019-10-08
add versioning API

ziyu-guo:
*Issue #, if available:*

*Description of changes:*

This implementation of get_version() API is designed on the assumption that TVM/Treelite portion of DLR package is the only part that needs explicit versioning, on top of the underlying TVM/Treelite runtime libs, since these libs are built into libdlr. 

Other executors like TFLite and 3rd party vendor runtime libs will each implement their own version API and hook it to python level get_version(), or leave it as Not Implemented. Therefore this API is expected to be used together with get_backend() to provide the whole picture.

Open to suggestions.

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -18,7 +18,7 @@ extern "C" { // Open extern "C" block
 #define DLR_MINOR 0
 /*! \brief patch version */
 #define DLR_PATCH 0
-/*! \brief mxnet version */
+/*! \brief DLR version */
 #define DLR_VERSION (DLR_MAJOR*10000 + DLR_MINOR*100 + DLR_PATCH)
 /*! \brief helper for making version number */
 #define DLR_MAKE_VERSION(major, minor, patch) ((major)*10000 + (minor)*100 + patch)
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -12,6 +12,17 @@ extern "C" { // Open extern "C" block
 #endif
 #endif // __cplusplus
 
+/*! \brief major version */
+#define DLR_MAJOR 1
+/*! \brief minor version */
+#define DLR_MINOR 0
+/*! \brief patch version */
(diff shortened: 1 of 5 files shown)


include/dlr.h
```diff
@@ -12,6 +12,17 @@ extern "C" { // Open extern "C" block
 #endif
 #endif // __cplusplus
 
+/*! \brief major version */
+#define DLR_MAJOR 1
+/*! \brief minor version */
+#define DLR_MINOR 0
+/*! \brief patch version */
+#define DLR_PATCH 0
+/*! \brief mxnet version */
```
zhiics: Why MXNet version?

zhiics:
What's the sample output of the current version? It looks like 10000?

ziyu-guo:
yes, e.g. 10500 means version 1.5 and 10501 means version 1.5.1

This is consistent with the mxnet_version attr at the end of mxnet .json files.

zhiics:
Yeah, I was wondering why not just print 1.5.1 directly.

include/dlr.h
```diff
@@ -12,6 +12,17 @@ extern "C" { // Open extern "C" block
...
 #endif // __cplusplus
 
+/*! \brief major version */
+#define DLR_MAJOR 1
+/*! \brief minor version */
+#define DLR_MINOR 0
+/*! \brief patch version */
+#define DLR_PATCH 0
+/*! \brief DLR version */
+#define DLR_VERSION (DLR_MAJOR*10000 + DLR_MINOR*100 + DLR_PATCH)
```
zhiics: Do we need this and the one below? If so, please change `GetDLRVersion` to reflect how you get the version `1.x.y`. Otherwise, we can just remove.

ziyu-guo: GetDLRVersion() outputs "1.x.y" string via 'out' arg now.

Issue #60, opened 2019-05-23
Load model .so and json/param from different locations


ziyu-guo:
Currently DLR assumes model artifact files are stored under the same folder. We need to be able to specify separate paths for the files.

PR #98, opened 2019-10-18
support loading tvm model files from different folders in C-API

ziyu-guo:
Customers raised issue that they'd like DLR to load artifacts from different locations in #60 

TVM runtime have separate LoadParams() API. However DLR uses filenames to select from different constructors. Invoking LoadParams() after DLR constructor means DLR constructor is no longer stateless, which is undesirable. 

This PR adds the ability to supply DLR C-API constructor with additional paths that contains the .json and .param file, by making CreateDLRModel() a variadic function. It allows all folders containing part of the model to be supplied in any order, in the following fashion:

`CreateDLRModel(&model_handle, path1, device_type, device_id, num_addl_paths, path2, path3)`

A macro in dlr.h handles the default value for additional arguments to maintain backward compatibility.


apivovarov:
The reason for that requirement is probably caused by the security restrictions in Android allowing app to load shared libraries only from app local folder. This is why users might want to keep `.so` file in app local folder while `.params` and `.json` might be in some shared folder.

DLR C API has the following method to construct DLR Model
```
# model_path Path to the top-level directory containing the model.
int CreateDLRModel(DLRModelHandle *handle,
                   const char *model_path,
                   int dev_type,
                   int dev_id);
```

we can also allow `model_path` to contain colon `:` separated list of files
e.g.
```
model_path=/data/local/dlr_mobilenet_v1_1.0_224/model.json:/data/local/dlr_mobilenet_v1_1.0_224/model.params:/data/user/0/org.tensorflow.lite.examples.classification/dlr_mobilenet_v1_1.0_224/model.so
```

Colon `:` is not a valid file name symbol in all OSs (Linux, Mac, Windows)

include/dlr.h
```diff
@@ -37,7 +41,7 @@ typedef void* DLRModelHandle;
 /*!
  \brief Creates a DLR model.
  \param handle The pointer to save the model handle.
- \param model_path Path to the top-level directory containing the model.
```
apivovarov: I guess what we discussed was that we still support old approach (model_path = top-level directory containing the model) and in addition to it we support model_path string in format of colon-separated list of paths.

I expect to see changes in the implementations of
```
DLRBackend dlr::GetBackend(const std::string& dirname)
ModelPath dlr::GetTvmPaths(const std::string& dirname) 
ModelPath dlr::GetTreelitePaths(const std::string& dirname)
```
These method should check if `dirname` contains `:`

if NOT then these methods should use old implementations which uses `ListDir` to convert `dirname` to `paths_vec` vector.

If YES then they should split dirname using `:` into `paths_vec` vector.

ziyu-guo: Production and consumption of path vector are separated so we only need to check for ':' in one centralized location, in CreateDLRModel(). 

Luckily ListDir() was able to handle multiple dirs in the first place, and both GetBackend() and GetTvmPaths() have been modified to accept path vector strings by looping on ListDir().

Thanks for catching the missing part in GetTreelitePaths(). I'll add corresponding change for consistency.

src/dlr_common.cc
```diff
@@ -45,14 +45,27 @@ void dlr::ListDir(const std::string& dirname, std::vector<std::string>& paths) {
   }
 }
 
-DLRBackend dlr::GetBackend(const std::string& dirname) {
+std::vector<std::string> dlr::split(const char* s, char delimiter) {
+   std::vector<std::string> tokens;
+   std::string token;
+   std::istringstream tokenStream(s);
+   while (std::getline(tokenStream, token, delimiter))
+   {
```
apivovarov: Fix the formatting

include/dlr.h
```diff
@@ -37,7 +37,8 @@ typedef void* DLRModelHandle;
...
+ \param model_path Directory containing the model files. 
+                   Colon-separated paths if multiple directories needed
```
apivovarov: Lets explicitly say that we support colon-separated list of folders (or files)
```
model_path Path to the folder containing the model files. Or colon-separated list of folders (or files) if model files are stored in multiple folders
```

include/dlr_common.h
```diff
@@ -42,6 +42,8 @@ typedef struct {
 
 void ListDir(const std::string& dirname, std::vector<std::string>& paths);
 
+std::vector<std::string> split(const char* s, char delimiter);
```
apivovarov: Can we use `split` function from `dmlc/common.h` https://github.com/dmlc/dmlc-core/blob/d06aceb92be200c9649258da726be6bb8e4d6556/include/dmlc/common.h#L23

```
namespace dmlc {
inline std::vector<std::string> Split(const std::string& s, char delim) {
```

PR #99, opened 2019-10-18
Save libs and bins relative to build folder

apivovarov:
This is needed for cases when we simultaneously build project for different architectures. For Example when building the project using Android Gradle.

For Example below is libdlr built for 4 Android architectures:
```
./dlr/build/intermediates/cmake/release/obj/armeabi-v7a/libdlr.so
./dlr/build/intermediates/cmake/release/obj/x86/libdlr.so
./dlr/build/intermediates/cmake/release/obj/arm64-v8a/libdlr.so
./dlr/build/intermediates/cmake/release/obj/x86_64/libdlr.so
```


ziyu-guo:
Good suggestion, can you change test and demo executable path to build/bin/ as well?

PR #105, opened 2019-11-02
add GetRuntimeEnabled()

ziyu-guo:
Adds get_enabled_devices() --> List in python API and GetDLRRuntimeEnabled() --> bool in C API. 

These functions queries underlying TVM runtime to determine compute devices turned-on for currently build. This is needed for correct error-handling when DLR parses metadata from Neo service.

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -216,6 +216,15 @@ int SetDLRNumThreads(DLRModelHandle* handle, int threads);
  */
 int UseDLRCPUAffinity(DLRModelHandle* handle, int use);
 
+/*!
+ \brief Check if a particular compute device is enabled for the runtime
+ \param handle The model handle returned from CreateDLRModel().
+ \param device The compute device type 
+ \param enabled Pointer to save bool result
+ \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+int GetDLRRuntimeEnabled(DLRModelHandle* handle, const char* device, bool* enabled);
+
 /*! \} */
 
 #ifdef __cplusplus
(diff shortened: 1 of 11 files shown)


include/dlr.h
```diff
@@ -216,6 +216,15 @@ int SetDLRNumThreads(DLRModelHandle* handle, int threads);
...
 int UseDLRCPUAffinity(DLRModelHandle* handle, int use);
 
+/*!
+ \brief Check if a particular compute device is enabled for the runtime
+ \param handle The model handle returned from CreateDLRModel().
+ \param device The compute device type 
+ \param enabled Pointer to save bool result
+ \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+int GetDLRRuntimeEnabled(DLRModelHandle* handle, const char* device, bool* enabled);
```
apivovarov: As discussed lets see if we can remove `handle` from `GetDLRRuntimeEnabled`.

`TVMModel::GetRuntimeEnabled` probably can be taken out from TVMModel class. Rename it to `GetTVMRuntimeEnabled` ?

ziyu-guo: Good point, but OTOH, Removing handle means we lose the executor info. That would defeat the whole effort of hiding TVMModel/TreeliteModel/TFLITEModel classes behind DLRModel() to create an uniform API.  Besides, other runtimes might have similar multi-device support, and we cannot afford to add a GetXXXRuntimeEnabled() API for every executor.

apivovarov: Egg and chicken problem. To create DLR model you need to know device name. To get list of enabled devices you need to have created DLR model.

Can we remove handle and and add backend parameter to the function?
```
int GetDLRRuntimeEnabled(const char* backend, const char* device, int* enabled);
```

ziyu-guo: Ok. To make c-api and python-api inconsistent, we'll need to refactor Python API and move get_version(), get_runtime_enabled() out of DLRModel() class. 

That also means we need to move _load_lib into maybe utils.py, so _LIB is available before DLRModel() ctor.

PR #109, opened 2019-11-19
TVM Python Support and Xilinx FPGA APIs (#1)

edelaye:
Changes TVM library to dynamic for python support
Adds Xilinx FPGA APIs as contrib

Note requires change to Jenkins build to install TVM python from 3rdparty/tvm/python (python setup.py ...)

### DIFF
--- a/python/dlr/contrib/xilinx_vai.py
+++ b/python/dlr/contrib/xilinx_vai.py
@@ -1,0 +1,82 @@
+"""
+All userspace code authored by Xilinx is released under the following license:
+Copyright (C) 2019 Xilinx, Inc
+Licensed under the Apache License, Version 2.0 (the "License"). You may
+not use this file except in compliance with the License. A copy of the
+License is located at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+License for the specific language governing permissions and limitations
+under the License.
+"""
+
+"""
+Registration of Xilinx Vitis-AI fused acceleration operation for aceleration of 
+convolutional neural networks
+
+
+This Vitis-AI acceleration exploits DPU hardware accelerator built for 
+following evaluation boards:
+- Ultra96: https://www.xilinx.com/products/boards-and-kits/1-vad4rl.html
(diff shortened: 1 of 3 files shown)


python/dlr/contrib/xilinx_vai.py
```diff
@@ -0,0 +1,89 @@
...
+import warnings
+import numpy as np
+
+try:
+    from dnndk import n2cube
+except:
+    warnings.warn("Could not import dnndk n2cube")
+
+
+@tvm.register_func("tvm.accel.accel_fused")
```
ziyu-guo: It seems this global is already registered in the TVM side, so DLR does not need to import TVM Python and repeat it here. How is it going to be used though? Are you planning to call this global from the run() function of a C++ adaptor ( e.g. XilinxModel class derived from DLRModel) to execute a compiled model on FPGA?

ziyu-guo:
Hi @edelaye, can you elaborate the general workflow to build and run DLR on FPGAs with your change? E.g new dependencies, cmdlines to build/install dependencies, and demo/test scripts to run a compiled model on device. Thanks!

python/dlr/contrib/xilinx_vai.py
jtuyls: Thank you for reviewing the PR. It's indeed already added on TVM side but we were not sure how often the TVM submodule in the Neo-ai-dlr is updated. The registration of this operation is basically the only adjustment  we need in the TVM runtime to be able to accelerate models on the FPGA using the DLR and that's why we added it here. It's implemented in python for generality, in the same way as we did it in the TVM PR. We believe it would be good to enable runtime python bindings (see changes in makefile) in addition to the C++ ones as they do in TVM, Tensorflow, PyTorch, ... because it makes it in general more convenient for accelerator developers to plug in their code and removes the need for recompilation.

For executing models, we just use the DLRModel class and the corresponding run function. All the FPGA stuff happens under the hood. Specifically, the directory that is pointed to should contain the three files retrieved from the TVM build module (lib, graph, params) and an additional library file that contains all information we need for execution on the FPGA. The TVM graph will contain a 'fused acceleration' operation which is registered and implemented here. If this operation is not registered, there will be an error in the (DLR) TVM runtime that the 'fused_accel' operation could not be found. When the DLRModel is executed, this operation will be called at its respective turn and it will use the provided accelerator lib file to run part of the model on the FPGA and return the results. (See attached diagram for the high-level flow).
xlnx-tvm-dlr-flow

PR #112, opened 2019-12-11
Prod countermgr

harshbafna:
*Issue #, if available:*

*Description of changes:*


dkasar-aws:
Call Home Feature:  A simple mechanism to update a global record counting the activation of the Neo runtime on every endpoint. The goal of this feature is to count the number of inferences and the number of active endpoints in deployment at any time with the minimal capture of data and minimal use of system and network resources by employing non-blocking functions.

python/dlr/counter/utils/dlrlogger.py
```diff
@@ -0,0 +1,8 @@
+import logging
+
+
+logging.basicConfig(filename='ccm_app.log', filemode='w', format='%(asctime)s-%(levelname)s-%(message)s')
```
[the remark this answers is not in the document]
dkasar-aws: We had removed this logging functionality and replaced it with exiting AI.DLR logging mechanism.

python/dlr/counter/config.py
```diff
@@ -0,0 +1,5 @@
+call_home_url = 'https://r4sir2jo37.execute-api.us-east-2.amazonaws.com/testing/'
```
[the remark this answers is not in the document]
dkasar-aws: Agree. We will update the url before the PR gets merged.

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class CallCounterMgr(object):
+    RUNTIME_LOAD = 1
+    MODEL_LOAD = 2
+    MODEL_RUN = 3
+    _instance = None
+
+    @staticmethod
+    def get_instance():
```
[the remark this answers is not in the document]
dkasar-aws: The notification text has configured and can printed before sending information to a server.

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            if not self.is_device_info_published():
+                pub_data = {'record_type': CallCounterMgr.RUNTIME_LOAD}
+                pub_data.update(self.system.get_device_info())
+                self._push(pub_data)
+        except Exception as e:
+            logger.exception("while dlr runtime load", exc_info=True)
+        return True
+
+    def model_count(self, count_type, model):
```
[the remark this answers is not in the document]
dkasar-aws: Required changes have incorporated in the PR. We kept configured duration to 5 secs.

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,144 @@
...
+    MODEL_LOAD = 2
+    MODEL_RUN = 3
+    _instance = None
+
+    @staticmethod
+    def get_instance():
+        """return single instance of class"""
+        if CallCounterMgr._instance is None:
+            if CallCounterMgr.is_feature_enabled():
+                print(config.CALL_HOME_USR_NOTIFICATION)
```
wuchih-amazon: I'd like this message to be part of setuptools entrypoint or post-installation. I'd like user being informed when they install DLR.

wuchih-amazon: If I understand this correctly, user needs to restart their application to load the static config file in order to shutdown the phone-home, correct?

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,144 @@
...
+                    data = json.load(ccm_json_file)
+                    if 'false' == str(data['ccm']).lower():
+                        return False
+                    else:
+                        return True
+            else:
+                return True
+        except Exception as e:
+            logging.exception("while in reading ccm config file")
+        return True
```
wuchih-amazon: I want this to be a static hard-coded variable, DEFAULT_ENABLE_FEATURE = True.

Reason being this value is yet decided, and I don't want the hardcoded True as return value

python/dlr/counter/model_metric.py
```diff
@@ -0,0 +1,66 @@
...
+        while ModelMetric._pub_model_metric:
+            time.sleep(config.CALL_HOME_MODEL_RUN_COUNT_TIME_SECS)
+            for key, val in ModelExecCounter.get_model_counts_dict().items():
+                self.model_run_info_publish(ModelMetric.MODEL_RUN, key, val)
+            ModelExecCounter.clear_model_counts()
+
+    def model_run_info_publish(self, model_event_type, model, count=0):
+        """push model load information at time model load time"""
+        try:
+            _md5model = hashlib.md5(model.encode())
```
wuchih-amazon: I'd like this to be a separate utility function. `def hash_model_name()`

python/dlr/counter/system.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self):
+        self._device = DeviceInfo()
+
+        try:
+            # retrieve device information
+            self._device.machine = platform.machine()
+            self._device.arch = platform.architecture()[0]
+            _uuid = ':'.join(
+                ['{:02x}'.format((uuid.getnode() >> ele) & 0xff) for ele in range(0, 8 * 6, 8)][::-1])
+            _md5uuid = hashlib.md5(_uuid.encode())
```
wuchih-amazon: same here, I'd suggest a separate hash helper for future auditing purpose.

python/dlr/counter/system.py
```diff
@@ -0,0 +1,110 @@
...
```
wuchih-amazon: seem like this is re-usable code here.

python/dlr/counter/config.py
```diff
@@ -0,0 +1,12 @@
+CALL_HOME_URL = 'https://r4sir2jo37.execute-api.us-east-2.amazonaws.com/testing/'
+CALL_HOME_RECORD_FILE = 'ccm_record.txt'
+CALL_HOME_USER_CONFIG_FILE = 'ccm_config.json'
+CALL_HOME_PUBLISH_MESSAGE_MAX_QUEUE_SIZE = 100
+CALL_HOME_MAX_WORKERS_THREADS = 5
+CALL_HOME_MODEL_RUN_COUNT_TIME_SECS = 5
+CALL_HOME_USR_NOTIFICATION = """\n CALL HOME FEATURE DISABLED
```
mnuyens: This is misleading, this sounds like it is already disabled, but it should be more like this is what you need to do to disable this feature

python/dlr/counter/model_metric.py
mnuyens: +1

python/dlr/counter/counter_mgr.py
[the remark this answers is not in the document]
mnuyens: 5 secs is still too short, that results in a data usage of over 500mb per month, and this is actually accumulating values right not just accumulating messages.

python/dlr/counter/system.py
dkasar-aws: + @mnuyens ,
We have moved hashlib functionality to new helper module.

python/dlr/counter/system.py
dkasar-aws: + @mnuyens ,
We have moved hashlib functionality to new helper module.

python/dlr/counter/counter_mgr.py
dkasar-aws: + @mnuyens,
We have added this message in the setup script and will inform user before installing the DLR.
Yes, the user needs to restart their application after place the required configuration (to disable the feature).

python/dlr/counter/config.py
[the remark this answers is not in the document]
dkasar-aws: @mnuyens,
Could you please provide production-ready URL.

python/dlr/counter/counter_mgr.py
dkasar-aws: Added static variable to hold default configuration of feature True.

python/dlr/counter/config.py
dkasar-aws: The notification text has corrected and moved to setup tool script.

python/dlr/counter/counter_mgr.py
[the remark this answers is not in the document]
dkasar-aws: Ok. We have modified wait time to 300 secs, that would takes less data usage approx. 8mb per month. Please let us know if you can suggest best configurable time here.

dkasar-aws:
> We need to just stop reporting metrics in an exponential backoff if we see repeated failures from the services.

@mnuyens 
We have added static counter to count repeated failure during reporting metrics data.
CALL_HOME_REQ_STOP_MAX_COUNT = 3, will check maximum request failure and then after stop sending data on server.

We here try to avoid additional dependency of python library.

python/dlr/counter/config.py
dkasar-aws: After modify user notification text message and incorporate into setup utility,
Below is python installation command shows the message ..
 
/neo-ai-dlr/python$ python3 setup.py install --user

 CALL HOME FEATURE ENABLED
                            
 Your device information and model load/execution count metric data being sent to a server.                             
 You can disable this feature by using below configuration steps.                             
	1. Create a config file, ccm_config.json in your application path.                             
	2. Added below format content in it,                             
		{
			"ccm" : "false"
		}                             
	3. Restart DLR application.
running install
running bdist_egg
running egg_info
...

python/setup.py
```diff
@@ -30,6 +30,17 @@
...
+#user data consent notification
+CALL_HOME_USR_NOTIFICATION = """\n CALL HOME FEATURE ENABLED
+                            \n Your device information and model load/execution count metric data being sent to a server. \
+                            \n You can disable this feature by using below configuration steps. \
+                            \n\t1. Create a config file, ccm_config.json in your application path. \
+                            \n\t2. Added below format content in it, \
+                            \n\t\t{\n\t\t\t"ccm" : "false"\n\t\t} \
+                            \n\t3. Restart DLR application."""
+
+print(CALL_HOME_USR_NOTIFICATION)
```
wuchih-amazon: @mnuyens @dkasar-aws,

This won't work for pip install unless verbose flag is enabled. As Max suggested, we should display this in application level when user import DLRModel. In other words, can we revise this back to the previous revision, where in counter_mgr.py
```
        if CallCounterMgr._instance is None:
            if CallCounterMgr.is_feature_enabled():
                 print(warning_msg)
```

python/setup.py
```diff
@@ -30,6 +30,17 @@
 if not LIB_PATH:
     raise RuntimeError('libdlr.so missing. Please compile first using CMake')
 
+#user data consent notification
+CALL_HOME_USR_NOTIFICATION = """\n CALL HOME FEATURE ENABLED
```
wuchih-amazon: this can move to config.py

python/setup.py
wuchih-amazon: Can we also move print after ```setup```? so the message show up at the bottom? in case user enable the verbose flag?

python/dlr/counter/counter_mgr.py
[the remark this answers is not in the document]
wuchih-amazon: Since pip prohibit message being printed in setup.py, we need to have this supported.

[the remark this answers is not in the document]
wuchih-amazon: @mnuyens  do you want this to be repeated whenever model is being loaded? or just when runtime being loaded?

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,144 @@
...
+        return CallCounterMgr._instance
+
+    def __init__(self):
+        try:
+            self.msg_publisher = MsgPublisher.get_instance()
+            machine_typ = platform.machine()
+            os_name = platform.system()
+            os_supt = "{0}_{1}".format(os_name, machine_typ)
+            self.system = Factory.get_system(os_supt)
+            self.model_metric = ModelMetric.get_instance(self.system.get_device_uuid())
```
wuchih-amazon: I just realize this while running unit tests. If system is not supported, don't activate counter. This will throw exception.

python/dlr/counter/model_metric.py
dkasar-aws: Already Done

python/dlr/counter/counter_mgr.py
dkasar-aws: Added unsupported system handling and will throw exception in the case.
CounterMgr will not activate in the case.

python/setup.py
dkasar-aws: We moved this notification to config.

dkasar-aws:
> 
> 
> Since pip install cannot properly display this message, we should display the same disclaimer in the README.md.

We have added the feature notification in README.md file. Please let us know correctness of format and content.

python/setup.py
dkasar-aws: @wuchih-amazon, @mnuyens 
As suggested above, the notification may not work in setup. We prefer to use notification in CounterMgr(as per earlier approach) and added the notification content in README.md file. This way we can resolve the issue and avoid CounterMgr module reference in Setup utils.

python/dlr/counter/model_metric.py
```diff
@@ -0,0 +1,70 @@
...
+                self.model_run_info_publish(ModelMetric.MODEL_RUN, key, val)
+            ModelExecCounter.clear_model_counts()
+
+    def model_run_info_publish(self, model_event_type, model, count=0):
+        """push model load information at time model load time"""
+        try:
+            _md5model = get_hash_string(model.encode())
+            _md5model = str(_md5model.hexdigest())
+            pub_data = {'record_type': model_event_type, 'model': _md5model, 'uuid': self.uuid, 'run_count': count}
+            time.sleep(config.CALL_HOME_MODEL_RUN_COUNT_TIME_SECS)
```
wuchih-amazon: I wrote an integration test for this branch, and discover that this will cause the program to hang.  Please do make sure that this thread is exit properly when program ends.

wuchih-amazon: and what's the exact reason why we have a timeout here?

wuchih-amazon:
I'm also seeing this sporadically whenever program exit the test.
`
Error in atexit._run_exitfuncs:
  File "/usr/local/lib/python3.6/dist-packages/dlr/counter/counter_mgr.py", line 148, in stop
    self.model_metric.stop()
  File "/usr/local/lib/python3.6/dist-packages/dlr/counter/model_metric.py", line 69, in stop
    for key, val in mod_dict.items():
`

python/dlr/counter/publisher.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            self.record_queue.put(data)
+            self.event.set()
+        except queue.Full as e:
+            logging.exception("ccm msg publisher queue full")
+        except Exception as e:
+            logging.exception("unable to record messages in queue", exc_info=True)
+
+    def _process_queue(self):
+        while self.event.wait() and not MsgPublisher._stop_processing:
```
wuchih-amazon: It takes me awhile, but I pinpoint down the thread was hanging because the deadlock here. 

I cannot determine whether it's because `self.event.wait` or `self.record_queue.get`; however this thread is still running after program exited. This need to be fixed. 

Again, it's reproducible with given `python3 -m pytest tests/python/unittest/test_counter_mgr.py` And you can see program won't stop the execution unless key interrupt exception.

dkasar-aws:
I am unable to reproduce exact the application hang issue in a development environment.
Currently, I am doing code analysis and adding necessary checks/care to avoid dead locks.
Also, I am fixing, send interrupt to the sleeping thread.

dkasar-aws:
I have added a few more checks, refined code changes to fix thread hang issues and replaced production beta URL in this feature.
Could you please check if still the issue reproducible with the latest DLR package.

If no issue found then please merge the pull request.

python/dlr/counter/publisher.py
```diff
@@ -34,25 +32,19 @@ def __init__(self):
...
             self.record_queue.put(data)
-            self.event.set()
         except queue.Full as e:
             logging.exception("ccm msg publisher queue full")
         except Exception as e:
             logging.exception("unable to record messages in queue", exc_info=True)
 
     def _process_queue(self):
-        while self.event.wait() and not MsgPublisher._stop_processing:
+        while True and not MsgPublisher._stop_processing:
```
wuchih-amazon: `True` is unnecessary

python/dlr/counter/publisher.py
```diff
@@ -34,25 +32,19 @@ def __init__(self):
...
             self.client.send(self.record_queue.get(block=False))
```
wuchih-amazon: `self.record_queue.get(block=False)`  will throw `empty exception` if block is set false.  This needs to be wrapped with try-catch in case of empty queue.

python/dlr/counter/model_metric.py
```diff
@@ -64,17 +64,13 @@ def push(self, data):
...
 
     def stop(self):
-        if self.executor:
-            with self.condition:
-                self.condition.notify_all()
+        with self.condition:
+            self.condition.notify_all()
             ModelMetric._pub_model_metric = False
             self.executor.shutdown(wait=False)
             mod_dict = ModelExecCounter.get_dict()
```
wuchih-amazon: Wouldn't `self.condition.notify_all()` notifies the thread in line 42? Why are we trying to publish dictionary items again?

wuchih-amazon:
> Request changes, dead block resolution might introduce unnecessary empty queue exception. On the positive side, the dead lock seems to be resolved.

I made a CR on what I think it's a better solution. Please review. https://github.com/dkasar-aws/neo-ai-dlr/pull/6/files

python/dlr/counter/model_exec_counter.py
```diff
@@ -1,12 +1,15 @@
...
 
 lock = threading.Lock()
 
 class ModelExecCounter(object):
     model_dict = {}
+    intr_dict = {}
     INIT = 1
     INCREMENT = 2
     CLEAR = 3
+    GETINTDICT = 4
```
wuchih-amazon: why are we having another enum?

python/dlr/counter/model_exec_counter.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @staticmethod
+    def update_dict(op, model=None, cnt=1):
+        with lock:
+            if op == ModelExecCounter.INIT:
+                ModelExecCounter.model_dict[str(model)] = 1
+            elif op == ModelExecCounter.INCREMENT:
+                ModelExecCounter.model_dict[str(model)] = cnt
+            elif op == ModelExecCounter.GETINTDICT:
+                ModelExecCounter.intr_dict = copy.deepcopy(ModelExecCounter.model_dict)
```
wuchih-amazon: I assume this is for dictionary size change exception, correct?

Instead of introducing a new enum, can we just copy a list of keys in model_metrics.py? https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,153 @@
...
+            logging.exception("unable to complete model count", exc_info=True)
+
+    def push(self, data):
+        """publish information to Server"""
+        if self.msg_publisher:
+            self.msg_publisher.send(json.dumps(data))
+
+    def stop(self):
+        with relinst:
+            if self.model_metric:
```
wuchih-amazon: There is a null pointer exception here when `if self.system is None:` in line 64.

Please fix this and add unit test for unsupported system as well

python/dlr/counter/model_exec_counter.py
dkasar-aws: I have considered below use case, 1. To unblock the main thread to continuous update primary dictionary, we created a temporary dictionary (using deep copy instead of shallow copy).
- This way the main thread will not wait until Publisher read items from the primary dictionary.
- Easy to share data between CounterMgr and Publisher

wuchih-amazon:
@dkasar-aws

I submit another PR for client side certificate for urllib3. That should take care the SSL handshake problem... (Again, that needs to be verified through Jenkin tests again). Also, Jenkin tests are timing out due to deadlock. Please review the commits where tests started to fail.

dkasar-aws:
I have reverted a few parts of the last commit to checking the deadlock issue resolved.
Currently, we are waiting for CI/CD build completion. 
It seems ci/cd build stuck at the same stage.

chjasonwu:
@dkasar-aws  the jenkin is back online, would you mind review the changes?

wuchih-amazon:
@dkasar-aws  jenkin is back online, mind review the tests?

dkasar-aws:
After merging master to feature branch, CI/CD build successful.

python/dlr/counter/publisher.py
```diff
@@ -0,0 +1,57 @@
...
+            self.record_queue.put(data, block=False)
+        except queue.Full as e:
+            logging.exception("ccm msg publisher queue full")
+        except Exception as e:
+            logging.exception("unable to record messages in queue", exc_info=True)
+
+    def _process_queue(self):
+        while not MsgPublisher._stop_processing:
+            if MsgPublisher.resp_cnt < config.CALL_HOME_REQ_STOP_MAX_COUNT:
+                resp_code = self.client.send(self.record_queue.get(block=True, timeout=2))
```
wuchih-amazon: Please add try catch for timeout exception

python/dlr/counter/publisher.py
```diff
@@ -0,0 +1,57 @@
...
+                if resp_code != 200:
+                    MsgPublisher.resp_cnt += 1
+            else:
+                self.record_queue.get(block=False)
+        logging.info("ccm msg publisher execution stopped")
+
+    def stop(self):
+        try:
+            self.record_queue.task_done()
```
wuchih-amazon: `ERROR:root:called task_done more than required times on ccm queue
  File "/home/ubuntu/.local/lib/python3.6/site-packages/dlr/counter/publisher.py", line 53, in stop
    self.record_queue.task_done()
  File "/usr/lib/python3.6/queue.py", line 68, in task_done
    raise ValueError('task_done() called too many times')
ValueError: task_done() called too many times`

Perhaps we should check the queue is empty prior calling task_done. This is because you have timeout implemented as well.

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,168 @@
...
+            return CallCounterMgr._instance
+
+    def __init__(self):
+        try:
+            machine_typ = platform.machine()
+            os_name = platform.system()
+            os_supt = "{0}_{1}".format(os_name, machine_typ)
+            self.system = Factory.get_system(os_supt)
+            if self.system is None:
+                raise Exception("unsupported system")
```
wuchih-amazon: This is a bug. We're shouldn't throw exception and stop DLR because we don't support phone-home. The reason is that your system validation will be expired someday in the future; however, that doesn't mean DLR won't work. Instead, we should log the warning and disable the feature. In addition, I suggest that we should have the validation check prior initialize CallCounterMgr.  Something similar to `is_feature_enabled`, we can have something like `is_feature_supported` and log the warning there.

Following is an exception to indicate this problem. I'm testing this on imx8 device with Linux_aarch64. DLR will work, but because of the version problem, DLR was forced to stop. 

[Errno 2] No such file or directory: '/opt/ml/errors/errors.log'
ERROR:root:while in counter mgr init
  File "/home/torizon/workplace/wuchih/neo-ai-dlr/python/dlr/counter/counter_mgr.py", line 72, in __init__
    raise Exception("unsupported system" + os_name + "," + os_supt)
Exception: unsupported systemLinux,Linux_aarch64
ERROR:root:unsupported system for call home feature
  File "/home/torizon/workplace/wuchih/neo-ai-dlr/python/dlr/counter/counter_mgr.py", line 55, in get_instance
    CallCounterMgr._instance = CallCounterMgr()
  File "/home/torizon/workplace/wuchih/neo-ai-dlr/python/dlr/counter/counter_mgr.py", line 81, in __init__
    raise e
  File "/home/torizon/workplace/wuchih/neo-ai-dlr/python/dlr/counter/counter_mgr.py", line 72, in __init__
    raise Exception("unsupported system" + os_name + "," + os_supt)
Exception: unsupported systemLinux,Linux_aarch64

python/dlr/counter/system.py
```diff
@@ -0,0 +1,109 @@
...
+system_list = ["Linux_ARM", "Linux_x86"]
+
+
+# factory class for System wrapper class
+class Factory:
+    @staticmethod
+    def get_system(sys_typ):
+        """Return instance of System as per operating system type"""
+        try:
+            map_sys_typ = [item for item in system_list if item.lower() in sys_typ.lower()]
```
wuchih-amazon: @dkasar-aws  I talked to @ziyu-guo. This system support validation is not necessary correct based on this https://github.com/neo-ai/neo-ai-dlr/blob/master/doc/install.rst#table-1-list-of-supported-devices

If I understand correctly, you're trying to separate implementation between Android and Linux, right?

I reached this problem while testing DLR on other linux distributions. Based on my conversation with other device engineers, linux distro shouldn't be a problem. 

If that's the case, please consider add support for all other linux distributions as well.

python/dlr/counter/system.py
```diff
@@ -0,0 +1,109 @@
...
+            self._device.arch = platform.architecture()[0]
+            _uuid = ':'.join(
+                ['{:02x}'.format((uuid.getnode() >> ele) & 0xff) for ele in range(0, 8 * 6, 8)][::-1])
+            _md5uuid = get_hash_string(_uuid.encode()) 
+            self._device.uuid = str(_md5uuid.hexdigest())
+            self._device.osname = platform.system()
+            dist = platform.dist()
+            self._device.dist = " ".join(x for x in dist)
+            self._device.name = platform.node()
+        except Exception as e:
```
wuchih-amazon: This can be consolidate with Linux_ARM logic.

python/dlr/counter/publisher.py
dkasar-aws: Added appropriate exception handling.

python/dlr/counter/publisher.py
dkasar-aws: Added few more checks. I am not able to reproduce the scenario.
Please let us know observation with latest changes.

python/dlr/counter/counter_mgr.py
dkasar-aws: We have just propagate the exception till Call Manger handler function but not intention to stop the application.
The logger by default print exception with trace info. 
We have disabled it now.
About Linux variants system, now we handle that scenario by using default System class Linux_ARM which may also retrieve required information.

python/dlr/counter/system.py
dkasar-aws: Assume all these devices on Linux OS, distributions. We have tested the feature on Linux 86, and ARM architecture machines.
Let us know to expand scope of the feature.

wuchih-amazon: We can expand this for all machines for linux. Phone-home shouldn't be architecture dependent. (at least not in python proportion)

python/dlr/counter/counter_mgr.py
```diff
@@ -0,0 +1,172 @@
...
+    def __init__(self):
+        try:
+            machine_typ = platform.machine()
+            os_name = platform.system()
+            os_supt = "{0}_{1}".format(os_name, machine_typ)
+            self.system = Factory.get_system(os_supt)
+            if self.system is None:
+                #support all kind of linux distributions
+                if os_name.lower() == 'linux':
+                    self.system = Factory.get_system("Linux_ARM")
```
wuchih-amazon: I think we should expand support to linux os, rather than hard-coded machine architecture here.

python/dlr/counter/counter_mgr.py
wuchih-amazon: > About Linux variants system, now we handle that scenario by using default System class Linux_ARM which may also retrieve required information.

I don't think that hard-coded `Linux_ARM` to replace all other machines is a sound approach, though `aarch64` is ARM chip in this case. There are too many variants to validate which machine will work for DLR and which don't.  However, in our case, we would only like to validate that this is linux based system, correct? Since phone-home is an application layer feature, I don't think any machine architecture matter in this case. Hence, I'd still suggest that we allow to send message for Linux base machine. And of course, we can always review the system validation later with @ziyu-guo

python/dlr/counter/counter_mgr.py
dkasar-aws: We have expanded support to Linux OS.

PR #127, opened 2020-01-26
Add TF C API adapter

apivovarov:
TF C API adapter allows DLR to run .pb files directly using Native Tensorflow C API (libtensorflow.so).

This is needed for 
* Alexa team to run .pb files on p3 instances in their C++ based product (gstreamer).
* Autodesk to run .pb file on aws lambda (unpacked lambda function size limit is 250 GB, libtensorflow.so is abut 200MB).

Pre-built Tensorflow for CPU and GPU can be found here https://www.tensorflow.org/install/lang_c

Better to build TF lib from source as described here https://github.com/tensorflow/tensorflow/blob/r1.15/tensorflow/tools/lib_package/README.md

TF lib for aws lambda: https://pivovaa-us-west-1.s3-us-west-1.amazonaws.com/libtensorflow-ivybridge.tar.gz

TF lib for p3 cuda 10.0 / cudnn 7.6.5: https://pivovaa-us-west-1.s3-us-west-1.amazonaws.com/libtensorflow-gpu-10-0.tar.gz

Command to build Tensorflow 1.15.0:
```
git clone https://github.com/tensorflow/tensorflow.git
cd tensorflow
git checkout r.1.15

./configure

bazel build -s \
--explain /tmp/build.log \
--verbose_explanations \
--config opt \
--config=v1 \
--verbose_failures \
--config=noaws \
--config=nogcp \
--config=nohdfs \
--config=noignite \
--config=nokafka \
--config=nonccl \
//tensorflow/tools/lib_package:libtensorflow
```
This command will produce `libtensorflow.tar.gz`.
Extract it to some folder and export TF_LIB to point to the folder.

Command to build DLR:
```
cmake -DWITH_TENSORFLOW_LIB=$TF_LIB -DCMAKE_BUILD_TYPE=Release ..

make -j$(nproc)
```

Command to run DLR/TF test `dlr_tensorflow_test`
```
LD_LIBRARY_PATH=$TF_LIB/lib ./dlr_tensorflow_test
```

### DIFF
--- a/include/dlr_tensorflow/dlr_tensorflow.h
+++ b/include/dlr_tensorflow/dlr_tensorflow.h
@@ -1,0 +1,64 @@
+#ifndef DLR_TENSORFLOW_H_
+#define DLR_TENSORFLOW_H_
+
+#include "dlr_common.h"
+#include "tensorflow/c/c_api.h"
+
+namespace dlr {
+
+/*! \brief Get the paths of the Tensorflow model files.
+ */
+std::string GetTensorflowFile(const std::string& dirname);
+
+/*! \brief free_buffer function used to cleanup memory after TF model is built.
+ */
+void FreeBuffer(void* data, size_t length);
+
+/*! \brief read tensorflow model file.
+ */
+TF_Buffer* ReadTFFile(const char* file);
+
+/*! \brief class TensorflowModel
+ */
+class TensorflowModel : public DLRModel {
+ private:
+  TF_Status* status_;
+  TF_Graph* graph_;
+  TF_Session* sess_;
+  // input_names_ are declared in base class
+  std::vector<std::string> output_names_;
+  std::vector<TF_Output> inputs_;
+  std::vector<TF_Output> outputs_;
+  std::vector<TF_Tensor*> input_tensors_;
+  std::vector<TF_Tensor*> output_tensors_;
+  void LoadFrozenModel(const char* pb_file);
+  void GenTensorSpec(bool is_input, const int batch_size);
+  int GetInputId(const char* name);
+
+ public:
+  /*! \brief Load model files from given folder path.
+   */
+  explicit TensorflowModel(const std::string& model_path, const DLContext& ctx,
+                           const std::vector<std::string>& inputs,
+                           const std::vector<std::string>& outputs,
+                           const int batch_size, const int threads);
+  ~TensorflowModel();
+
+  virtual const char* GetInputName(int index) const override;
+  virtual const char* GetWeightName(int index) const override;
+  virtual std::vector<std::string> GetWeightNames() const override;
+  virtual void GetInput(const char* name, float* input) override;
+  virtual void SetInput(const char* name, const int64_t* shape, float* input,
+                        int dim) override;
+  virtual void Run() override;
+  virtual void GetOutput(int index, float* out) override;
+  virtual void GetOutputShape(int index, int64_t* shape) const override;
+  virtual void GetOutputSizeDim(int index, int64_t* size, int* dim) override;
+  virtual const char* GetBackend() const override;
+  virtual void SetNumThreads(int threads) override;
+  virtual void UseCPUAffinity(bool use) override;
+};
+
+}  // namespace dlr
+
+#endif  // DLR_TENSORFLOW_H_
--- a/src/dlr.cc
+++ b/src/dlr.cc
@@ -6,6 +6,9 @@
 #ifdef DLR_TFLITE
 #include "dlr_tflite/dlr_tflite.h"
 #endif // DLR_TFLITE
+#ifdef DLR_TENSORFLOW
+#include "dlr_tensorflow/dlr_tensorflow.h"
+#endif // DLR_TENSORFLOW
 
 #include <locale>
 
@@ -128,6 +131,30 @@ int CreateDLRModelFromTFLite(DLRModelHandle *handle,
   API_END();
 }
 #endif // DLR_TFLITE
+
+#ifdef DLR_TENSORFLOW
+/*! \brief Translate c args from ctypes to std types for DLRModelFromTensorflow ctor.
+ */
+int CreateDLRModelFromTensorflow(DLRModelHandle *handle,
+                                 const char *model_path,
+                                 const char* inputs[], int input_size,
+                                 const char* outputs[], int output_size,
+                                 const int batch_size,
+                                 int threads) {
+  API_BEGIN();
+  const std::string model_path_string(model_path);
+  // TensorflowModel class does not use DLContext internally
+  DLContext ctx;
+  ctx.device_type = static_cast<DLDeviceType>(1); // 1 - kDLCPU
+  ctx.device_id = 0;
+  std::vector<std::string> v_inputs(inputs, inputs + input_size);
+  std::vector<std::string> v_outputs(outputs, outputs + output_size);
+  DLRModel* model = new TensorflowModel(
+      model_path_string, ctx, v_inputs, v_outputs, batch_size, threads);
+  *handle = model;
+  API_END();
+}
+#endif // DLR_TENSORFLOW
 
 /*! \brief Translate c args from ctypes to std types for DLRModel ctor.
  */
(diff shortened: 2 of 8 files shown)


include/dlr_tensorflow/dlr_tensorflow.h
```diff
@@ -0,0 +1,64 @@
...
+  virtual void GetOutputSizeDim(int index, int64_t* size, int* dim) override;
+  virtual const char* GetBackend() const override;
+  virtual void SetNumThreads(int threads) override;
+  virtual void UseCPUAffinity(bool use) override;
+};
+
+} // namespace dlr
+
+
+#endif  // DLR_TENSORFLOW_H_
```
trevor-m: Add newline at EOF

include/dlr_tensorflow/dlr_tensorflow.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+void FreeBuffer(void* data, size_t length);
+
+/*! \brief read tensorflow model file.
+ */
+TF_Buffer* ReadTFFile(const char* file);
+
+/*! \brief class TensorflowModel
+ */
+class TensorflowModel: public DLRModel {
```
trevor-m: `class TensorflowModel : public DLRModel`
Add space
Use clang-format on all modified files

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -0,0 +1,290 @@
...
+    int64_t dims[n_dim];
+    TF_GraphGetTensorShape(graph_, oper_out, dims, n_dim, status_);
+    if (TF_GetCode(status_) != TF_OK) {
+      LOG(FATAL) << "ERROR: TF_GraphGetTensorShape failed " << TF_Message(status_);
+      return; // unreachable
+    }
+    if (dims[0] == -1) {
+      dims[0] = 1;
+    }
+    size_t num_pixels = dims[0];
```
trevor-m: Should we use `int64_t` instead of `size_t` since `size_t` could be 32 bits?

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -0,0 +1,290 @@
...
+
+  LoadFrozenModel(pb_file.c_str());
+
+  // Using assignment operator to copy one vector to other
+  input_names_ = inputs;
+  output_names_ = outputs;
+  num_inputs_ = inputs.size();
+  num_outputs_ = outputs.size();
+
+  GenTensorSpec(true);
```
trevor-m: To make this more clear
```
GenTensorSpec(/*is_input=*/true);
GenTensorSpec(/*is_input=*/false);
```

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -0,0 +1,290 @@
...
+  if (TF_GetCode(status_) != TF_OK) {
+    LOG(FATAL) << "ERROR: Unable to import graph " << TF_Message(status_);
+    return; // unreachable
+  }
+  LOG(INFO) << "Successfully imported graph";
+  TF_DeleteImportGraphDefOptions(opts);
+  TF_DeleteBuffer(graph_def);
+}
+
+void TensorflowModel::GenTensorSpec(bool isInput) {
```
trevor-m: `isInput` -> `is_input`

src/dlr.cc
```diff
@@ -129,6 +132,29 @@ int CreateDLRModelFromTFLite(DLRModelHandle *handle,
 }
 #endif // DLR_TFLITE
 
+#ifdef DLR_TENSORFLOW
+/*! \brief Translate c args from ctypes to std types for DLRModelFromTensorflow ctor.
+ */
+int CreateDLRModelFromTensorflow(DLRModelHandle *handle,
+                   const char *model_path,
```
trevor-m: Allignment is off here. Use `clang-format --style=google -i src/dlr.cc`

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -0,0 +1,290 @@
...
+}
+
+
+// Constructor
+TensorflowModel::TensorflowModel(const std::string& model_path,
+                                 const DLContext& ctx,
+                                 const std::vector<std::string>& inputs,
+                                 const std::vector<std::string>& outputs,
+                                 const int threads
+                                ): DLRModel(ctx, DLRBackend::kTENSORFLOW) {
```
trevor-m: Use clang-format

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -0,0 +1,290 @@
...
+      LOG(FATAL) << "ERROR: TF_GraphGetTensorNumDims failed " << TF_Message(status_);
+      return; // unreachable
+    }
+    int64_t dims[n_dim];
+    TF_GraphGetTensorShape(graph_, oper_out, dims, n_dim, status_);
+    if (TF_GetCode(status_) != TF_OK) {
+      LOG(FATAL) << "ERROR: TF_GraphGetTensorShape failed " << TF_Message(status_);
+      return; // unreachable
+    }
+    if (dims[0] == -1) {
```
trevor-m: This will override the batch size to only allow batch size = 1 when the model has a dynamic batch dimension. If we want to allow dynamic shapes to work correctly, we need to defer `TF_AllocateTensor` until `SetInput` is called.

If we don't care about dynamic shapes and always want to override the batch dimension to 1 in those cases, we should change the dimension in the graph using `TF_GraphSetTensorShape`. This should allow TF to further optimize the graph by using static shapes.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.h#L199-L219

src/dlr_tensorflow/dlr_tensorflow.cc
trevor-m: Lets use `num_elements` instead of `num_pixels` since we aren't always using images.

trevor-m: n_dim can be 0 when the tensor is a scalar. So `num_elements` should be initialized to 1 and the following loop should start at dim 0.

Scalars:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api_test.cc#L440-L447

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -0,0 +1,290 @@
...
+      return; // unreachable
+    }
+
+    TF_Operation* op = TF_GraphOperationByName(graph_, op_name.c_str());
+    if (op == NULL) {
+      LOG(FATAL) << "ERROR: inputOp TF_GraphOperationByName failed for operation " << op_name;
+      return; // unreachable
+    }
+    TF_Output oper_out = {op, op_out_id};
+    int n_dim = TF_GraphGetTensorNumDims(graph_, oper_out, status_);
```
trevor-m: `const int`

include/dlr_tensorflow/dlr_tensorflow.h
apivovarov: added "Ensure line feed at file end on Save setting" ide

src/dlr.cc
apivovarov: Let me fix this particular block and run `clang-format` for existing old files in a separate PR to make it clear what was actually changed.

src/dlr_tensorflow/dlr_tensorflow.cc
apivovarov: Since TF_AllocateTensor needs `size_t` for 'len` we decided to use `size_t` for `num_elements`
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/tf_tensor.h#L91

src/dlr_tensorflow/dlr_tensorflow.cc
apivovarov: TVM compiled model has fixed batch size. DLR API was designed with this assumption.
Since compiled model has fixed batch size and DRL is designed for that we should ask about batch size in TensorflowModel constructor and call `TF_GraphSetTensorShape` in the initialization phase (before setInput)

PR #130, opened 2020-01-29
Ask input shape in TF C API

apivovarov:
TF models might have dynamic H and W.
In DLR we allocate input tensor once and reuse it for entire model life.
To make dynamic shape models work in DLR we should ask about input shape at the construction time.

Example:
```
const int64_t dims[4] = {batch_size, 224, 224, 3};
const DLR_TFTensorDesc inputs[1] = {{"input:0", dims, 4}};
const char* outputs[1] = {"MobilenetV1/Predictions/Reshape_1:0"};
const int threads = -1; // undefined

CreateDLRModelFromTensorflow(&handle, "model.pb", inputs, 1, outputs, 1, threads);
```

### DIFF
--- a/src/dlr_tensorflow/dlr_tensorflow.cc
+++ b/src/dlr_tensorflow/dlr_tensorflow.cc
@@ -70,73 +70,62 @@ void TensorflowModel::LoadFrozenModel(const char* pb_file) {
   TF_DeleteBuffer(graph_def);
 }
 
-void TensorflowModel::GenTensorSpec(bool is_input, const int batch_size) {
-  std::vector<std::string> tensor_names =
-      is_input ? input_names_ : output_names_;
+TF_Output TensorflowModel::ParseTensorName(const std::string& t_name) {
   std::regex r("^(.+):(\\d+)$");
-  for (std::string t_name : tensor_names) {
-    std::smatch match;
-    std::string op_name;
-    int op_out_id;
-    if (std::regex_search(t_name, match, r)) {
-      op_name = match.str(1);
-      op_out_id = std::stoi(match.str(2));
-    } else {
-      LOG(FATAL) << "ERROR: failed to parse tensor name " << t_name;
-      return;  // unreachable
-    }
+  std::smatch match;
+  if (!std::regex_search(t_name, match, r)) {
+    LOG(FATAL) << "ERROR: failed to parse tensor name " << t_name;
+  }
+  std::string op_name = match.str(1);
+  int op_out_id = std::stoi(match.str(2));
+  TF_Operation* op = TF_GraphOperationByName(graph_, op_name.c_str());
+  if (op == NULL) {
+    LOG(FATAL) << "ERROR: TF_GraphOperationByName failed for operation "
+               << op_name;
+  }
+  TF_Output oper_out = {op, op_out_id};
+  return oper_out;
+}
 
-    TF_Operation* op = TF_GraphOperationByName(graph_, op_name.c_str());
-    if (op == NULL) {
-      LOG(FATAL)
-          << "ERROR: inputOp TF_GraphOperationByName failed for operation "
-          << op_name;
-      return;  // unreachable
-    }
-    TF_Output oper_out = {op, op_out_id};
-    const int n_dim = TF_GraphGetTensorNumDims(graph_, oper_out, status_);
-    if (TF_GetCode(status_) != TF_OK) {
-      LOG(FATAL) << "ERROR: TF_GraphGetTensorNumDims failed "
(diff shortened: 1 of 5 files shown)


src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -199,6 +215,18 @@ TensorflowModel::TensorflowModel(const std::string& model_path,
   }
   TF_DeleteSessionOptions(opts);
   LOG(INFO) << "Tensorflow Session was created";
+
+  // Run inference to allocate output Tensors and calculate output shapes.
+  for (int i = 0; i < num_inputs_; i++) {
+    TF_Tensor* tensor = input_tensors_[i];
+    int64_t num_elements = TF_TensorElementCount(tensor);
+    float* in_t_data = (float*)TF_TensorData(tensor);
+    for (int i = 0; i < num_elements; i++) {
```
trevor-m: Lets use `std::fill` here instead of the loop.

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -70,73 +72,90 @@ void TensorflowModel::LoadFrozenModel(const char* pb_file) {
...
+  }
+  TF_Output oper_out = {op, op_out_id};
+  return oper_out;
+}
+
+void TensorflowModel::PrepInputs() {
+  for (int i = 0; i < num_inputs_; i++) {
+    std::string t_name = input_names_[i];
+    TF_Output oper_out = ParseTensorName(t_name);
+    std::vector<int64_t> shape = input_shapes_[i];
```
trevor-m: Make shape a const reference.

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -70,73 +72,90 @@ void TensorflowModel::LoadFrozenModel(const char* pb_file) {
...
-      if (TF_GetCode(status_) != TF_OK) {
-        LOG(FATAL) << "ERROR: TF_GraphSetTensorShape failed "
-                   << TF_Message(status_);
-        return;  // unreachable
-      }
-    }
+
     size_t num_elements = 1;
     for (int z = 0; z < n_dim; z++) {
-      if (dims[z] == -1) {
```
trevor-m: Should we keep this check in case user gives a bad input shape containing -1?

apivovarov: good catch!

src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -70,73 +70,62 @@ void TensorflowModel::LoadFrozenModel(const char* pb_file) {
...
-        LOG(FATAL) << "ERROR: TF_GraphSetTensorShape failed "
-                   << TF_Message(status_);
-        return;  // unreachable
-      }
-    }
+void TensorflowModel::PrepInputs() {
+  for (int i = 0; i < num_inputs_; i++) {
+    const std::string t_name = input_names_[i];
+    const TF_Output oper_out = ParseTensorName(t_name);
+    const std::vector<int64_t> shape = input_shapes_[i];
```
trevor-m: Make shape a reference so it doesn't have to be copied

apivovarov: ok, fixed it in 3 places
```
diff --git a/src/dlr_tensorflow/dlr_tensorflow.cc b/src/dlr_tensorflow/dlr_tensorflow.cc
index b1bcb64..bae473e 100644
--- a/src/dlr_tensorflow/dlr_tensorflow.cc
+++ b/src/dlr_tensorflow/dlr_tensorflow.cc
@@ -89,9 +89,9 @@ TF_Output TensorflowModel::ParseTensorName(const std::string& t_name) {
 
 void TensorflowModel::PrepInputs() {
   for (int i = 0; i < num_inputs_; i++) {
-    const std::string t_name = input_names_[i];
+    const std::string& t_name = input_names_[i];
     const TF_Output oper_out = ParseTensorName(t_name);
-    const std::vector<int64_t> shape = input_shapes_[i];
+    const std::vector<int64_t>& shape = input_shapes_[i];
     const int64_t* dims = shape.data();
     const int n_dim = shape.size();
     size_t num_elements = 1;
@@ -120,7 +120,7 @@ void TensorflowModel::PrepInputs() {
 }
 
 void TensorflowModel::PrepOutputs() {
-  for (std::string t_name : output_names_) {
+  for (std::string& t_name : output_names_) {
     TF_Output oper_out = ParseTensorName(t_name);
 
     outputs_.push_back(oper_out);
```

PR #133, opened 2020-02-04
[TF C API] Autodetect inputs and outputs

apivovarov:
Autodetect inputs and outputs in TF Frozen Graph (`.pb` file).

Now users can use Generic `CreateDLRModel` function to create DLR model from `.pb` file, e.g.
```
  const char* model_file =
      "./mobilenet_v1_1.0_224/mobilenet_v1_1.0_224_frozen.pb";
  const int dev_type = 1;  // 1 - kDLCPU
  const int dev_id = 0;

  DLRModelHandle handle;
  if (CreateDLRModel(&handle, model_file, dev_type, dev_id)) {
    FAIL() << DLRGetLastError() << std::endl;
  }
  LOG(INFO) << "CreateDLRModel - OK";
```

I experimented with different `.pb` files (some of them had lots of garbage tensors) to find the best approach to find "real" graph outputs. After all I decided to use the following algo to detect the outputs.

* If at least one output of operator "A" is used by other operators then ignore all outputs of operator "A"
* If all outputs of operator "A" do not have consumers then add all outputs of operator "A" to the list of outputs.

### DIFF
--- a/src/dlr_tensorflow/dlr_tensorflow.cc
+++ b/src/dlr_tensorflow/dlr_tensorflow.cc
@@ -87,6 +87,102 @@ TF_Output TensorflowModel::ParseTensorName(const std::string& t_name) {
   return oper_out;
 }
 
+void TensorflowModel::DetectInputs() {
+  size_t pos = 0;
+  TF_Operation* op;
+  while ((op = TF_GraphNextOperation(graph_, &pos)) != nullptr) {
+    const std::string op_type = TF_OperationOpType(op);
+    const int n_in = TF_OperationNumInputs(op);
+    const int n_out = TF_OperationNumOutputs(op);
+    const std::string op_name = TF_OperationName(op);
+    if (op_type == "Placeholder" && n_in == 0 && n_out == 1) {
+      input_names_.push_back(op_name + ":0");
+    }
+  }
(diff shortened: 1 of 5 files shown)


src/dlr_tensorflow/dlr_tensorflow.cc
```diff
@@ -87,6 +87,84 @@ TF_Output TensorflowModel::ParseTensorName(const std::string& t_name) {
...
+  TF_Operation* op;
+  // while loop
+  while ((op = TF_GraphNextOperation(graph_, &pos)) != nullptr) {
+    const std::string op_type = TF_OperationOpType(op);
+    const int n_out = TF_OperationNumOutputs(op);
+    const int n_cout = TF_OperationNumControlOutputs(op);
+    const std::string op_name = TF_OperationName(op);
+    if (op_type != "Const" && op_type != "Assign" && op_type != "NoOp" &&
+        op_type != "Placeholder" && n_cout == 0) {
+      int n_cons = 0;
```
trevor-m: Since there are already a few `n_*` variables going around here, it may be helpful to change `n_cons` to `n_consumers` to be more clear.

apivovarov: fixed

PR #140, opened 2020-02-08
Fix error logger for container and DLR

ashishgupta023:
- Add a new parameter on DLRModel for logger
- By default, logs to sys.stdout
- Inference containers write to the error log path defined by Sagemaker

What has been tested?
- built DLR from source on Linux and it does not print anything on import
- verified the logs are written to file and printed to console
- verified batch transforms on sagemaker with containers on trusted neo service accounts and containers on other AWS accounts


container/dockerd-entrypoint.sh
```diff
@@ -6,6 +6,7 @@ source /home/model-server/mms_config.sh
 
 if [[ "$1" = "serve" ]]; then
     shift 1
+    mkdir -p /opt/ml/errors
```
ashishgupta023: Sagemaker only mounts the error volumes for trusted service accounts in which case it will always be there, for containers in customer accounts we need to create this folder

apivovarov:
Can you build and install dlr on Linux. Then open python 3 and import dlr.
Does it work? What it prints?

python/dlr/neologger.py
```diff
@@ -1,12 +1,7 @@
 """Logging API"""
 import logging
 
-'''
-Trusted error logs are mounted by default on the algorithm container's volume. 
-Any logs written to /opt/ml/errors/errors.log will be reported back to sagemaker.
-DO NOT CHANGE THE LOG FILE PATH/NAME
-'''
-def create_logger(log_file="/opt/ml/errors/errors.log", log_level=logging.DEBUG, verbose=True):
+def create_logger(log_file=None, log_level=logging.DEBUG, verbose=True):
```
apivovarov: Since log_file is optional, maybe put it as the last parameter.
`verbose` is confusing. . Lets find better name. Maybe `log_to_console=True/False`

python/dlr/api.py
```diff
@@ -7,13 +7,6 @@
 import logging
```
ziyu-guo: Can we remove this import and replace all remaining logging() reference in the file with neo_logger?

apivovarov:
Still interested to see what happened after "import dlr" on regular computer now. Can you try it and post the output to the description?

python/dlr/api.py
ashishgupta023: Done

python/dlr/neologger.py
ashishgupta023: Done

ashishgupta023:
image

ashishgupta023:
@apivovarov Attached the screenshot after building from source and importing. Also updated the description on what other tests were conducted.

apivovarov:
Nice! Thank you!

PR #142, opened 2020-02-13
Fix typo so that 400 error is thrown correctly

ashishgupta023:
This commit by Philip had a typo due to which error was still showing Internal Server Error 

https://github.com/neo-ai/neo-ai-dlr/commit/2895a6747f48dd7a420f0a0ac227236272af246b

Testing
- Batch Transform now shows ClientError: See job logs for more information


apivovarov:
nice catch

PR #146, opened 2020-02-19
[Tensorflow] Add DLR_TFConfig

apivovarov:
Tensorflow adapter changes / fixes:
1. Fix GPU memory leak in case `RunDLRModel` (TF backend) is invoked multiple times in loop.
2. Allows users to set Tensorflow `GPUOptions` such as `allow_growth` and `per_process_gpu_memory_fraction`. User needs to use `CreateDLRModelFromTensorflow` function to provide custom `DLR_TFConfig`.
3. If user uses generic `CreateDLRModel` then we set `allow_growth=True` and `per_process_gpu_memory_fraction=10%`. It allows effectively share GPU memory and run several DLR/TF_Session instances on one GPU. 

I tested Tensorflow models With and Without (`per_process_gpu_memory_fraction=10%` `allow_growth=True`) - No performance degradation was noticed in case we run one model on GPU.

If I run multiple models on GPU in parallel I observed gradual performance drop as the number of models increases, e.g.
```
# number of models  - avg inference time
1 - 8ms
2 - 13ms
4 - 22ms
...
14 - 84ms
```
Without that settings we can not run multiple models on GPU simultaneously because the first model allocates all GPU memory. As a result the second model initialization fails with `CUDA_ERROR_OUT_OF_MEMORY`

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -65,8 +65,26 @@ int CreateDLRModelFromTFLite(DLRModelHandle* handle, const char* model_path,
 #endif  // DLR_TFLITE
 
 #ifdef DLR_TENSORFLOW
+
 /*!
- \brief Input Tensor Description structure for CreateDLRModelFromTensorflow.
+ \brief Tensorflow GPU Options structure for Tensorflow Config structure.
+ */
+typedef struct DLR_TFGPUOptions {
+  int allow_growth;
+  double per_process_gpu_memory_fraction;
+} DLR_TFGPUOptions;
+
+/*!
+ \brief Tensorflow Config structure for CreateDLRModelFromTensorflow.
+ */
+typedef struct DLR_TFConfig {
+  int intra_op_parallelism_threads;
+  int inter_op_parallelism_threads;
+  DLR_TFGPUOptions gpu_options;
+} DLR_TFConfig;
+
+/*!
+ \brief Tensor Description structure for CreateDLRModelFromTensorflow.
  */
 typedef struct DLR_TFTensorDesc {
   const char* name;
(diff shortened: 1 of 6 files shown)


trevor-m:
Have you tried setting only `allow_growth=True` and not setting `per_process_gpu_memory_fraction`?

apivovarov:
I tried that.  Yes, it allocates less than 100% of the memory. But it still allocates more than it really needs.
For example, I have a style transformation model which “automatically" allocates 6GB of GPU memory (if I only say `allow_growth=True`)
But if I say `per_process_gpu_memory_fraction=0.01` then it starts with 350MB and quickly grows to 980MB and stays stable at that level.
The performance of the model is the same as when it uses 6GB.
So, I decided to use smth in between of 100% and 1%. 
“Generic” `CreateDLRModel` function uses 10% as a starting point. if model needs more it quickly allocates more memory.

include/dlr_tensorflow/dlr_tensorflow.h
```diff
@@ -48,7 +54,7 @@ class TensorflowModel : public DLRModel {
       const std::string& model_path, const DLContext& ctx,
       const std::vector<std::string>& inputs,
       const std::vector<std::vector<int64_t>>& input_shapes,
-      const std::vector<std::string>& outputs, const int threads);
+      const std::vector<std::string>& outputs, const DLR_TFConfig tf_config);
```
trevor-m: `const DLR_TFConfig tf_config` -> `const DLR_TFConfig& tf_config`

include/dlr.h
```diff
@@ -88,7 +106,7 @@ typedef struct DLR_TFTensorDesc {
 int CreateDLRModelFromTensorflow(DLRModelHandle* handle, const char* model_path,
                                  const DLR_TFTensorDesc* inputs, int input_size,
                                  const char* outputs[], int output_size,
-                                 const int threads);
+                                 const DLR_TFConfig tf_config);
```
trevor-m: `const DLR_TFConfig tf_config` -> `const DLR_TFConfig& tf_config`

include/dlr_tensorflow/dlr_tensorflow.h
apivovarov: fixed

include/dlr.h
apivovarov: Since it is C header we can pass by value or use pointer.

PR #150, opened 2020-03-03
[Docs] Improve install DLR steps

trevor-m:
Make steps for CPU, GPU, or OpenCL clearer. Also, include option of `-DUSE_TENSORRT=ON` for system installs of TensorRT.

Add `--force` to install of python package so that old builds of DLR are replaced.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -95,13 +95,12 @@ To build, create a subdirectory ``build``:
   cd build
   
 Building for CPU
+  Invoke CMake to generate a Makefile and then run GNU Make to compile:
 
-Invoke CMake to generate a Makefile and then run GNU Make to compile:
+  .. code-block:: bash
 
-.. code-block:: bash
-
-  cmake ..
-  make -j4         # Use 4 cores to compile sources in parallel
+    cmake ..
+    make -j4         # Use 4 cores to compile sources in parallel
 
 Building for GPU
   By default, DLR will be built with CPU support only. To enable support for NVIDIA GPUs, enable CUDA, CUDNN, and TensorRT by calling CMake with these extra options.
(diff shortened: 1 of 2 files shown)


doc/install.rst
```diff
@@ -109,19 +125,25 @@ By default, DLR will be built with CPU support only. To enable support for NVIDI
...
+Install Python package
+""""""""""""""""""""""
+
 Once the compilation is completed, install the Python package by running ``setup.py``:
 
 .. code-block:: bash
 
   cd ../python
-  python3 setup.py install --user
+  python3 setup.py install --user --force
```
apivovarov: Can you give an example of the case when -f is needed? (with material evidences)

trevor-m: If a previous install of DLR already exists, you will need to use force otherwise it will not replace the existing one. This is because the version of DLR is still 1.0 so it thinks it’s the same version.

apivovarov: Actually, it will replace it (this is why I asked for material evidences). The version is analyzed and the installation is skipped in case you use `pip` and your package needs to be downloaded from repo.
But `setup.py` does not analyze the version if the package files are on the drive.

Below are install log which shows installation and reinstallation without force. After the reinstallation all files were replaced with new one (exactly the dame size but updated time)
https://gist.githubusercontent.com/apivovarov/47715d2a804de7757388d29f95084534/raw/ed5870d6b1f91fd4e21f1141a147055ea5862276/install.log

apivovarov: [March 17, 2020, 1:27 PM] Morris, Trevor: Thanks, I am fine with removing force since I am unable to reproduce the behavior I saw before

PR #161, opened 2020-04-01
Add Hexagon adapter

apivovarov:
This PR adds DLR adapter to run models which use Hexagon NNLib and which were compiled by Hexagon SDK (for Linux or Android)

Currently we have a [hexagon-converter](https://github.com/apivovarov/hexagon-converter) which converts TFLite quantized models to C program which uses Hexagon NNLib API to assemble model graph and store model weights.
Hexagon SDK compiles the program to shared library for ARM cpu with particular extern functions, such as `dlr_hexagon_model_init`, `dlr_hexagon_model_exec`, etc.
Hexagon SDK also compiles Hexagon NNLib to shared library for Hexagon DSP.
The compilation output is
- `dlr_hexagon_model.so` - model graph and weights (Compiled for ARM CPU Linux/Android)
- `libhexagon_nn_skel.so` - Hexagon NNLib compiled for Hexagon DSP

DLR Adapter loads model file (e.g. `dlr_hexagon_model.so`) at runtime, it also "dlsym" extern functions (symbols) such as `dlr_hexagon_model_init`, `dlr_hexagon_model_exec` and calls them to init the model, get pointers to input/output buffers and run inference.

Hexagon SDK runtime uses `ADSP_LIBRARY_PATH` to find `libhexagon_nn_skel.so` file.
if `libhexagon_nn_skel.so` is present in the model folder then DLR adapter will set `ADSP_LIBRARY_PATH` automatically. If not then User needs to set `ADSP_LIBRARY_PATH` to point to `libhexagon_nn_skel.so` file folder

# Test
I tested DLR with hexagon compiled model `mobilenet_v1_0.75_224_quant_hexagon_model.so`.
I run it on `QCS605` dev board (Android 8.1, 64 bits)
Run log: https://textuploader.com/145zt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,6 +325,10 @@ if(WITH_TENSORFLOW_LIB)
     list(APPEND DLR_SRC "src/dlr_tensorflow/dlr_tensorflow.cc")
     list(APPEND DLR_LINKER_LIBS -L${WITH_TENSORFLOW_LIB}/lib -ltensorflow_framework -ltensorflow)
 endif()
+if(WITH_HEXAGON)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDLR_HEXAGON")
+    list(APPEND DLR_SRC "src/dlr_hexagon/dlr_hexagon.cc")
+endif()
 if(AAR_BUILD)
     list(APPEND DLR_SRC "src/jni/dlr_jni.cc")
 endif()
@@ -382,7 +386,7 @@ endforeach()
 add_custom_target(demo DEPENDS ${DEMO_EXECS})
 
 # Tests
-if(NOT(ANDROID_BUILD OR AAR_BUILD))
+if(NOT(AAR_BUILD))
   include(cmake/googletest.cmake)
   fetch_googletest(
     ${PROJECT_SOURCE_DIR}/cmake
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -382,7 +386,6 @@ endforeach()
 add_custom_target(demo DEPENDS ${DEMO_EXECS})
 
 # Tests
-if(NOT(ANDROID_BUILD OR AAR_BUILD))
```
apivovarov, line 385: In order to test Tensorflow, TFLite and Hexagon adapters on Android we need to enable GTest programs build for Android as well.

tests/cpp/dlr_hexagon/dlr_hexagon_test.cc
```diff
@@ -0,0 +1,201 @@
...
+
+  // Load image
+  size_t img_size = 224 * 224 * 3;
+  uint8_t* img = LoadImageAndPreprocess("cat224-3.txt", img_size);
+  LOG(INFO) << "Input sample: [" << +img[0] << "," << +img[1] << "..."
+            << +img[img_size - 1] << "]";
+
+  // SetDLRInput
+  const int64_t in_shape[4] = {1, 224, 224, 3};
+  if (SetDLRInput(&handle, input_name, in_shape, (float*)img, 4)) {
```
apivovarov, line 118: DLR API uses `float*` pointer to get starting address of image buffer. We still can pass starting address of uint8_t image buffer to DLR by casting `uint8_t*` to `float*`. Pointer size does not depend of data type. It just contains memory address - 64 or 32 bits depending on the platform.  It is ok to cast pointers in this case.

CMakeLists.txt
ziyu-guo: On a separate note, is it better to disable gtest by default? Maybe put it in a 'test' target and exclude it from 'all', since it adds quite some build time and the repos might not be accessible from some edge devices.

apivovarov: What you think if we GNU toolchain to build

doc/install.rst
```diff
@@ -145,7 +145,7 @@ Once the compilation is completed, install the Python package by running ``setup
 .. code-block:: bash
 
   cd ../python
-  python3 setup.py install --user --force
+  python3 setup.py install --user
```
apivovarov, line 148: Removal of `--force` was discussed here https://github.com/neo-ai/neo-ai-dlr/pull/150

CMakeLists.txt
ziyu-guo: I think we're already able to do that? E.g. I was able to compile dlr_tflite_test for imx8 with aarch64 crosscompiler designated in a toolchain.cmake.

apivovarov: Great!. Lets add ARM Linux cross-compilation to `install.rst`

PR #168, opened 2020-04-24
PhoneHome python wrapper

wuchih-amazon:

We'll emit metrics for model load and and device info for now. 

Performance test is conducted against m5.xlarge instance. With phone-home enabled, the inference is around 46ms, and without is around 45ms. There is no latency detected.


trevor-m:
Thanks @wuchih-amazon 
I have some questions about this change:
1. Is the call home syncronous or asynchronous?
2. Does having @call_home_lite on DLRModel.run incur a performance penalty?
3. Have we tested what happens on a device with no internet access? Will it still be able to load/run a model? Will there be long delays on each call while waiting for the connection to time out?

container/Dockerfile.cpu
```diff
@@ -17,7 +17,7 @@ RUN apt-get update && \
     && curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py && rm get-pip.py \
     && rm -rf /root/.cache/pip
 
-RUN pip3 install -U pip setuptools wheel
+RUN pip3 install -U pip setuptools wheel urllib3 certifi
```
cnv1989, line 20: Can we add these dependencies to setup.py instead?

wuchih-amazon: I actually remove dependencies for those.

python/dlr/counter/system.py
```diff
@@ -0,0 +1,217 @@
...
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_device_uuid(self):
+        """Return DeviceInfo uuid"""
+        raise NotImplementedError
+
+
+# Wrapper class
+class ARM(System):
```
trevor-m, line 29: Is it necessary to have all of these subclasses for each type of system? Is there really so much of a difference between the platform information from python apis between each type of system? It looks like most implementations of `System` are the same.

Can we just always collect all of the available data from `platform` and let the backend handle of the parsing and processing?

python/meta.py
```diff
@@ -0,0 +1,3 @@
+NAME = ['DLRModel']
```
trevor-m, line 1: This is a duplicate of `python/dlr/metadata.py`.

python/dlr/__init__.py
```diff
@@ -6,7 +6,8 @@
 from __future__ import absolute_import as _abs
 
 from .api import DLRModel
+from .metadata import NAME, VERSION
 
-__all__ = ['DLRModel']
+__all__ = NAME
 
-__version__ = "1.3.0"
+__version__ = VERSION
```
trevor-m, line 13: Why move this?

python/test/test_phone_home.py
```diff
@@ -0,0 +1,87 @@
...
+    def tearDown(self):
+        self.clean_config()
+
+    def clean_config(self):
+        config_path = PhoneHome.get_config_path()
+        if os.path.isfile(config_path):
+            os.remove(config_path)
+
+    def mock_func(self):
+        func = MagicMock(return_vaule=MagicMock())
```
trevor-m, line 20: return_value typo

python/meta.py
wuchih-amazon: removed.

python/dlr/__init__.py
wuchih-amazon: This is because of cyclical dependency conflicts when setup tool calling __init__.py, which then invoke singleton phone-home that requires setup tool to install the third party dependencies. This is a way to bypass that

python/test/test_phone_home.py
wuchih-amazon: that's embarrassing. lol fixed

python/dlr/counter/system.py
wuchih-amazon: I agree. This was written with different system support at first. But I have same remark with the original author. Let me revise this and get back to you.

trevor-m (approved):

PR #171, opened 2020-05-08
Implemented APIs to read from metadata file

cnv1989:

### Notes
NeoCompilation service ships `compiled.meta` file that contains metadata about the model. Metadata file has the following json format. The PR contains enables DLR to read ouput information from the metadata file.

This PR implements the following c apis

* `GetDLRHasMetadata`
* `GetDLROutputName`
* `GetDLROutputIndex`
* `GetDLROutputByName`

These apis are also exposed in python bindings as part for DLRModel class.

### Metadata Json
```json
{
    "Requirements": {
        "TargetDevice": "ML_M4"
    },
    "Compilation": {
        "CreatedTime": "2020-05-08T20:37:58.065747"
    },
    "Model": {
        "Inputs": [
            {
                "name": "input_tensor",
                "dtype": "float32",
                "shape": [
                    64,
                    224,
                    224,
                    3
                ]
            }
        ],
        "Outputs": [
            {
                "name": "ArgMax",
                "dtype": "int64",
                "shape": [
                    64
                ]
            },
            {
                "name": "softmax_tensor",
                "dtype": "float32",
                "shape": [
                    64,
                    1001
                ]
            }
        ]
    }
}
```
## ModelPeeker

```bash
[18:07:23] /home/DLR/neo-ai-dlr/src/dlr_tvm.cc:68: Loading metadata file: /home/DLR/compiled/compiled.meta
backend is tvm
num_inputs = 1
num_weights = 108
num_outputs = 2
input_names: input_tensor,
input_types: float32,
output shapes:
[64, ]
[64, 1001, ]
output_names: ArgMax (index: 0), softmax_tensor (index: 1),
output_types: int32, float32,
root@c60dadd939b7:/home/DLR/neo-ai-dlr/build# mv /home/DLR/compiled/compiled.meta /home/.
root@c60dadd939b7:/home/DLR/neo-ai-dlr/build# ./bin/model_peeker /home/DLR/compiled
[18:07:48] /home/DLR/neo-ai-dlr/src/dlr_tvm.cc:71: No metadata found
backend is tvm
num_inputs = 1
num_weights = 108
num_outputs = 2
input_names: input_tensor,
input_types: float32,
output shapes:
[64, ]
[64, 1001, ]
output_types: int32, float32,
```
### Without metadata 
```python
from dlr import DLRModel
```

```python
dlm = DLRModel("./compiled")
```

```python
dlm.get_output_dtypes()
```

    ['int32', 'float32']

```python
dlm.get_input_dtypes()
```

    ['float32']

```python
dlm.get_output_names()
```

    2020-05-20 18:16:20,677 ERROR error in getting output names DLRModelImpl Model has no metadata!
      File "/home/DLR/neo-ai-dlr/python/dlr/api.py", line 110, in get_output_names
        return self._impl.get_output_names()
      File "/home/DLR/neo-ai-dlr/python/dlr/dlr_model.py", line 202, in get_output_names
        raise DLRError("Model has no metadata!")
    dlr.dlr_model.DLRError: Model has no metadata!

    ---------------------------------------------------------------------------

    DLRError                                  Traceback (most recent call last)

    <ipython-input-5-4ab75d921cbc> in <module>
    ----> 1 dlm.get_output_names()
    

    /home/DLR/neo-ai-dlr/python/dlr/api.py in get_output_names(self)
        111         except Exception as ex:
        112             self.neo_logger.exception("error in getting output names {} {}".format(self._impl.__class__.__name__, ex))
    --> 113             raise ex
        114 
        115     def get_version(self):

    /home/DLR/neo-ai-dlr/python/dlr/api.py in get_output_names(self)
        108     def get_output_names(self):
        109         try:
    --> 110             return self._impl.get_output_names()
        111         except Exception as ex:
        112             self.neo_logger.exception("error in getting output names {} {}".format(self._impl.__class__.__name__, ex))

    /home/DLR/neo-ai-dlr/python/dlr/dlr_model.py in get_output_names(self)
        200     def get_output_names(self):
        201         if not self.has_metadata():
    --> 202             raise DLRError("Model has no metadata!")
        203         else:
        204             return self.output_names

    DLRError: Model has no metadata!

```python

```

### With Metadata
```python
from dlr import DLRModel
```

```python
dlm = DLRModel("./compiled")
```

```python
dlm.get_output_dtypes()
```

    ['int32', 'float32']

```python
dlm.get_input_dtypes()
```

    ['float32']

```python
dlm.get_output_names()
```

    ['ArgMax', 'softmax_tensor']

```python
dlm.get_input_names()
```

    ['input_tensor']

```python

```


src/dlr_tvm.cc
```diff
@@ -198,3 +207,70 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+  } 
+  *index = idx;
+}
+
+void TVMModel::GetOuputByName(const char* name, float* out) {
+  int* output_index;
+  this->GetOutputIndex(name, output_index);
+  if (*output_index >= 0) {
+    this->GetOutput(*output_index, out);
+  } {
```
cnv1989, line 273: typo

.gitmodules
```diff
@@ -5,3 +5,7 @@
 [submodule "3rdparty/treelite"]
 	path = 3rdparty/treelite
 	url = https://github.com/neo-ai/treelite
+[submodule "3rdparty/json"]
+	path = 3rdparty/json
+	url = https://github.com/nlohmann/json.git
```
anijain2305, line 10: Just wondering if there are any licensing issues using this new `nlohmann/json.git` git directory?

trevor-m: According to some google searches the MIT license used by nlohmann is compatible with the Apache License 2.0. But we should check internally to be safe.

src/dlr.cc
```diff
@@ -105,6 +106,59 @@ extern "C" int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs) {
...
+extern "C" int GetDLRInputIndex(DLRModelHandle* handle, char* name, int* index) {
+  API_BEGIN();
+  TVMModel* model = static_cast<TVMModel*>(*handle);
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  CHECK(model->HasMetadata()) << "model has no metadata";
+  model->GetInputIndex(name, index);
+  API_END();
+}
+
+extern "C" int GetDLROuputByName(DLRModelHandle* handle, char* name, float* out) {
```
apivovarov, line 153: typo - `Ouput`

src/dlr_tvm.cc
```diff
@@ -31,6 +31,8 @@ ModelPath dlr::GetTvmPaths(std::vector<std::string> dirname) {
       paths.params = filename;
     } else if (filename == "version.json") {
```
cnv1989, line 32: BTW this comparison doesn't work in C++ I think. Left it as is for now. I should probably fix it.

.gitmodules
cnv1989: MIT License is pre-approved.

https://inside.amazon.com/en/services/legal/us/OpenSource/Pages/BlessedOpenSourceLicenses.aspx

src/dlr_tvm.cc
apivovarov: in Java it does not work. In C++ it works. http://www.cplusplus.com/reference/string/string/operators/

apivovarov:
Could you add the description of the PR?

src/dlr.cc
```diff
@@ -105,6 +106,77 @@ extern "C" int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs) {
...
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  *has_metadata = model->HasMetadata();
+  API_END();
+}
+
+extern "C" int GetDLRInputNames(DLRModelHandle* handle, char*** names) {
+  API_BEGIN();
+  TVMModel* model = static_cast<TVMModel*>(*handle);
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  CHECK(model->HasMetadata()) << "model has no metadata";
```
apivovarov, line 121: TFLite, TF backends do not need Metadata to get list of model InputNames. TVM backend does not need it as well

src/dlr.cc
```diff
@@ -105,6 +106,77 @@ extern "C" int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs) {
...
+extern "C" int GetDLRInputIndex(DLRModelHandle* handle, char* name, int* index) {
+  API_BEGIN();
+  TVMModel* model = static_cast<TVMModel*>(*handle);
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  CHECK(model->HasMetadata()) << "model has no metadata";
+  model->GetInputIndex(name, index);
+  API_END();
+}
+
+extern "C" int GetDLROutputByName(DLRModelHandle* handle, char* name, float* out) {
```
apivovarov, line 153: Lets use generic type `void*` for out

src/dlr.cc
```diff
@@ -105,6 +106,77 @@ extern "C" int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs) {
...
+  CHECK(model->HasMetadata()) << "model has no metadata";
+  model->GetOuputByName(name, out);
+  API_END();
+}
+
+extern "C" int GetDLRInputDataTypes(DLRModelHandle* handle, char*** dtypes) {
+  API_BEGIN();
+  TVMModel* model = static_cast<TVMModel*>(*handle);
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  CHECK(model->HasMetadata()) << "model has no metadata";
```
apivovarov, line 166: TFLite, TF backends do not need neo Metadata file to get model InputDataTypes.

src/dlr_common.cc
```diff
@@ -1,7 +1,7 @@
-#include "dlr_common.h"
```
apivovarov, line 1: why this change is needed?

src/dlr_tvm.cc
```diff
@@ -198,3 +207,90 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+void TVMModel::GetOuputDataTypes(char**& dtypes) const {
+  dtypes = (char**) malloc(this->num_outputs_);
+  std::string dtype;
+
+  for(int i = 0; i < this->num_outputs_; i++) {
+    dtype = this->metadata["Model"]["Outputs"][i]["dtype"];
+    dtypes[i] = (char *)malloc((dtype.size() + 1)*sizeof(char));
+    strcpy(dtypes[i], dtype.c_str());
+  }
+}
```
apivovarov, line 296: Unix txt file should end with End of line.

src/dlr_tvm.cc
```diff
@@ -198,3 +207,90 @@ void TVMModel::UseCPUAffinity(bool use) {
     LOG(INFO) << "CPU Affinity is disabled";
   }
 }
+
+bool TVMModel::HasMetadata() const {
+  if (this->metadata != nullptr) {
```
apivovarov, line 212: just return it

src/dlr_tvm.cc
```diff
@@ -198,3 +207,90 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+  return false;
+}
+
+void TVMModel::GetOutputNames(char**& names) const {
+  names = (char**) malloc(this->num_outputs_);
+  std::string name;
+
+  for(int i = 0; i < this->num_outputs_; i++) {
+    name = this->metadata["Model"]["Outputs"][i]["name"];
+    names[i] = (char *)malloc((name.size() + 1)*sizeof(char));
```
apivovarov, line 224: Lets not do malloc each time when this method is called. We can do the following. TVMModel internally has `vector<string> input_names` storing input names. Lets create another vector in TVMModel - `vector<const char*> c_input_names` which stores pointers to c_strings of input_names vectors. When user calls `GetOutputNames` we simply give user `c_input_names` pointer. Example: https://gist.github.com/apivovarov/10be0d6a12dfc294ac87ee287e70deab
Benefits of that solution that we do not call `malloc` on every single method call. Users do not need to do memory cleanup. Price for that solution is new vector which stores 64 bit addresses of actual input names - very cheap.

src/dlr_common.cc
cnv1989: not needed something my editor did i guess. Shouldn't cause any harm

src/dlr.cc
cnv1989: Synced offline. Will be implemented as a separate PR.

src/dlr_tvm.cc
cnv1989: 👍

src/dlr_tvm.cc
cnv1989: 👍

src/dlr.cc
cnv1989: Can you get it using tvm graph runtime?

@ashishgupta023 You could perhaps use that approach instead.

src/dlr.cc
apivovarov: Lets use new PR to fix existing methods. But for new methods we can use void*

src/dlr_tvm.cc
cnv1989: Cool. Nice thanks

src/dlr.cc
ashishgupta023: @apivovarov there is a public `get_input` function in DLR which hardcodes input data type to float32 which isnt't right I guess . https://github.com/neo-ai/neo-ai-dlr/blob/master/python/dlr/dlr_model.py#L356. We can use the metadata file to set input type for that or delete the function `get_input` if its not required ?

src/dlr_tvm.cc
```diff
@@ -198,3 +207,90 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+
+void TVMModel::GetOuputByName(const char* name, float* out) {
+  int* output_index;
+  this->GetOutputIndex(name, output_index);
+  if (*output_index >= 0) {
+    this->GetOutput(*output_index, out);
+  }
+}
+
+void TVMModel::GetInputDataTypes(char**& dtypes) const {
```
apivovarov, line 276: You can use `GetInputType(index)` to get input type from compiled model json. `GetInputType(index)` was added to neo-ai/tvm graph_runtime https://github.com/neo-ai/tvm/commits/r20200204 (update 3rdparty/tvm commit in neo-ai-dlr)

src/dlr.cc
apivovarov: Python DLR `get_input` method will use DLR C API `GetDLRInputDataTypes` to get model inputs types instead of using hardcoded `float32`.
Internally `GetDLRInputDataTypes` will use `DLRModel::GetInputDataTypes`. DLRModel is abstract class, each impl (TF, TFLite, TVM) will use its own approach to impl `GetInputDataTypes`. `TensorflowModel` and `TFLiteModel` will use TF API and TFLite API to query `.pb` and `.tflite` files to get inputs name/type. `TVMModel` will use tvm graph_runtime functions - `GetInputType(index)` (just added) and `GetInputName(index)`.

src/dlr_tvm.cc
trevor-m: In that case we should merge https://github.com/neo-ai/neo-ai-dlr/pull/174 first since latest neo-ai/tvm requires other changes in that PR for CI and Cmake

src/dlr.cc
cnv1989: Discussed this offline. The nodes names are not consistent with uncompiled model node names. We should be reading output related data from metadata and input related data from graph_runtime.

include/dlr.h
```diff
@@ -250,6 +251,46 @@ int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs);
...
+ error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL int GetDLROutputByName(DLRModelHandle* handle, char* name, float* out);
+
+/*!
+ \brief Gets the ouput node dtypes from the metadata file
+ \param handle The model handle returned from CreateDLRModel().
+ \param dtypes The pointer to save array containing output node dtypes.
+*/
+DLR_DLL int GetDLROuputDataType(DLRModelHandle* handle, const int index, char** dtype);
```
apivovarov, line 292: `GetDLROutputType` was added in PR176

include/dlr.h
```diff
@@ -250,6 +251,46 @@ int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs);
...
+*/
+DLR_DLL int GetDLROutputName(DLRModelHandle* handle, const int index, char** name);
+
+/*!
+ \brief Gets the ouput node index using the node name
+  \param handle The model handle returned from CreateDLRModel().
+  \param name The pointer pointing to the ouput node name.
+  \param index The pointer to save the corresponding index of the output node.
+*/
+DLR_DLL int GetDLROutputIndex(DLRModelHandle* handle, char* name, int* index);
```
apivovarov, line 275: `const char* name`

include/dlr.h
```diff
@@ -250,6 +251,46 @@ int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs);
...
+
+/*!
+ \brief Gets the output of the node of the given name from the model.
+ \param handle The model handle returned from CreateDLRModel().
+ \param name The name of the ouput node.
+ \param out The pointer to save the output data. This should be a pointer to an
+ array of size "size" from GetDLROutputSizeDim(). \return 0 for success, -1 for
+ error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL int GetDLROutputByName(DLRModelHandle* handle, char* name, float* out);
```
apivovarov, line 285: `const char* name`

.gitmodules
apivovarov: Could we just add single `nlohmann/json.hpp` file to our include dir?

include/dlr_common.h
```diff
@@ -108,6 +112,27 @@ class DLRModel {
   virtual const char* GetBackend() const = 0;
   virtual void SetNumThreads(int threads) = 0;
   virtual void UseCPUAffinity(bool use) = 0;
+
+  virtual bool HasMetadata() const {
+    return false;
+  };
```
apivovarov, line 118: remove `;`

.gitmodules
trevor-m: I found that TVM already has a library "picojson" we could use instead https://github.com/neo-ai/tvm/tree/dev/3rdparty/picojson

include/dlr_common.h
```diff
@@ -108,6 +112,27 @@ class DLRModel {
...
+
+  virtual void GetOutputName(const int index, char*& name) const {
```
apivovarov, line 120: `virtual const char* GetOutputName(int index) const`

include/dlr_common.h
```diff
@@ -108,6 +112,27 @@ class DLRModel {
...
+    LOG(FATAL) << "GetOutputName is not supported yet!";
+  };
+
+  virtual int GetOutputIndex(const char* name) const {
+    LOG(FATAL) << "GetOutputIndex is not supported yet!";
+    return -1;
+  };
+
+  virtual void GetOuputByName(const char* name, float* out){
```
apivovarov, line 129: missing space btw `){`. Better to use `clang-format -i --style=Google <file>`

include/dlr.h
apivovarov: Lets start using `void* out` for data - Other (existing) functions can be fixed in a separate PR

python/dlr/dlr_model.py
```diff
@@ -120,15 +120,19 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0):
         self.num_inputs = self._get_num_inputs()
         self.num_weights = self._get_num_weights()
         self.input_names = []
+        self.output_names = []
         self.weight_names = []
         self.input_shapes = {}   # Remember shape used in _set_input()
-        for i in range(self.num_inputs):
-            self.input_names.append(self._get_input_name(i))
+        self.output_dtypes = []
```
apivovarov, line 126: Lets add `input_dtypes` as well. `GetDLRInputType` was added to C API recently.

python/dlr/dlr_model.py
```diff
@@ -160,15 +164,42 @@ def get_input_names(self):
...
+                _check_call(_LIB.GetDLROutputName(byref(self.handle), i, byref(name)))
+                self.output_names.append(name.value.decode('utf-8'))
+
+    def _fetch_input_names(self):
+        for i in range(self.num_inputs):
+            self.input_names.append(self._get_input_name(i))
+        
+    def _fetch_output_dtypes(self):
+        self.output_dtypes = []
+        if self.has_metadata():
```
apivovarov, line 186: Metadata is needed only fo fetch Original Output names

src/dlr.cc
```diff
@@ -1,3 +1,4 @@
+#include <typeinfo>
```
apivovarov, line 1: Looks like it is not needed

src/dlr_common.cc
apivovarov: We use `cmake-format --style=Google` to format C/C++ files for consistency.

src/dlr_common.cc
```diff
@@ -53,6 +53,11 @@ void dlr::ListDir(const std::string& dirname, std::vector<std::string>& paths) {
   }
 }
 
+void dlr::LoadJsonFromFile(const std::string& path, nlohmann::json& jsonObject) {
+  std::ifstream jsonFile (path);
+  jsonFile >> jsonObject;
+};
```
apivovarov, line 59: remove `;`

src/dlr_tvm.cc
```diff
@@ -198,3 +207,40 @@ void TVMModel::UseCPUAffinity(bool use) {
     LOG(INFO) << "CPU Affinity is disabled";
   }
 }
+
+bool TVMModel::HasMetadata() const {
+  return this->metadata != nullptr;
+}
+
+
```
apivovarov, line 215: 2 empty lines

src/dlr.cc
apivovarov: Lets read OutputType and InputType from model.json (`GetDLRInputType` and `GetDLROutputType` already added to dlr.h). Since  `model.json` does not have Original model OutputName we have to read it from metadata file. So, the only field that we need to read from metadata file is Original model OutputName.

.gitmodules
apivovarov: Nice! BTW, This PR should be rebased to use the latests `3rdparty/tvm`

python/dlr/dlr_model.py
cnv1989: 👍

python/dlr/dlr_model.py
cnv1989: Yes. Is that a question of does this require a change?

include/dlr_common.h
cnv1989: 👍 Its definitely more C++ like.

include/dlr_common.h
cnv1989: 👍

apivovarov:
Can you add new C API usage to `demo/cpp/model_peeker.cc` and post its output (for model with/without metadata file) to PR description?

include/dlr.h
cnv1989: I would rather not change his here. I think we should keep it consistent with `virtual void GetOutput(int index, float* out) override;`

include/dlr.h
cnv1989: 👍

.gitmodules
cnv1989: By looking at picojson it seems more complicate to use compared to nlohmann. What do you think?  Also the https://github.com/nlohmann/json/commits/develop is begin contributed lot more actively.

cnv1989: > Could we just add single `nlohmann/json.hpp` file to our include dir?

We are adding a single_include directory. I am not sure what you mean?

src/dlr_tvm.cc
cnv1989: We could wait on input types related changes and release them in a separate PR. I would rather release outpu related changes  with DLR 1.2 @trevor-m @apivovarov

include/dlr.h
cnv1989: @apivovarov I think we should read output related data from the metadata file for tvm for the sake of consistency.

CMakeLists.txt
```diff
@@ -91,6 +92,7 @@ include_directories("${FMT_SRC}/include")
 include_directories("${TREELITE_SRC}/include")
 include_directories("${TREELITE_SRC}/runtime/native/include")
 include_directories("${PROJECT_SOURCE_DIR}/include")
+include_directories("${JSON_SRC}/single_include")
```
cnv1989, line 95: @apivovarov  we are adding a single_include here. Do you mean instead of cloning the entire repo we just clone this one file?

include/dlr.h
apivovarov: `const char* name`

include/dlr.h
```diff
@@ -274,6 +275,38 @@ DLR_DLL
...
+*/
+DLR_DLL int GetDLROutputName(DLRModelHandle* handle, const int index, const char** name);
+
+/*!
+ \brief Gets the ouput node index using the node name
+  \param handle The model handle returned from CreateDLRModel().
+  \param name The pointer pointing to the ouput node name.
+  \param index The pointer to save the corresponding index of the output node.
+*/
+DLR_DLL int GetDLROutputIndex(DLRModelHandle* handle, char* name, int* index);
```
apivovarov, line 298: `const char* name`

include/dlr_tvm.h
```diff
@@ -3,6 +3,7 @@
 
 #include <graph/graph_runtime.h>
 #include <tvm/runtime/memory.h>
+#include <json.hpp>
```
apivovarov, line 6: Documentation recommends to include it the following way `#include <nlohmann/json.hpp>` https://github.com/nlohmann/json#integration.

python/dlr/api.py
```diff
@@ -118,3 +118,16 @@ def get_version(self):
         except Exception as ex:
             self.neo_logger.exception("error in getting version {} {}".format(self._impl.__class__.__name__, ex))
             raise ex
+
+    def has_metadata(self):
+        try:
+            return self._impl.has_metadata()
+        except Exception as ex:
+            self.neo_logger.exception("error in checking for metadata file {} {}".format(self._impl.__class__.__name__, ex))
```
apivovarov, line 126: Do we need `raise ex` after this line?

demo/cpp/model_peeker.cc
```diff
@@ -42,6 +43,13 @@ void peek_model(DLRModelHandle model) {
     std::cout << input_names[i] << ", ";
   }
   std::cout << std::endl;
+  output_names.resize(num_outputs);
+  std::cout << "output_names: ";
+  for (int i = 0; i < num_outputs; i++) {
+    GetDLROutputName(&model, i, &output_names[i]);
+    std::cout << output_names[i] << ", ";
+  }
+  std::cout << std::endl;
```
apivovarov, line 52: Lets move it after `input_types:`, before `output shapes:`. Also, we should print `output_names` only if matadata is present.

include/dlr_tvm.h
apivovarov: BTW, `json.hpp` is included to `dlr_common.h` already.

include/dlr_common.h
```diff
@@ -5,6 +5,7 @@
 #include <dmlc/logging.h>
 #include <runtime_base.h>
 #include <sys/types.h>
+#include <json.hpp>
```
apivovarov, line 8: Documentation recommends to include it the following way #include <nlohmann/json.hpp> https://github.com/nlohmann/json#integration.

CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ endif()
 set(DLR_LINKER_LIBS "")
 
 set(TVM_SRC "${PROJECT_SOURCE_DIR}/3rdparty/tvm")
+set(JSON_SRC "${PROJECT_SOURCE_DIR}/3rdparty/nlohmann")
```
apivovarov, line 81: `json.hpp` is not compiled to `.o` file but included to other `.cc` files. So, we do not need to add it to `SRC`

include/dlr_common.h
```diff
@@ -111,6 +115,23 @@ class DLRModel {
...
+    return false;
+  }
+
+  virtual const char* GetOutputName(const int index) const {
+    LOG(FATAL) << "GetOutputName is not supported yet!";
+  }
+
+  virtual int GetOutputIndex(const char* name) const {
+    LOG(FATAL) << "GetOutputIndex is not supported yet!";
+    return -1;
```
apivovarov, line 129: Lets add comment `// unreachable`. `return -1 will never be executed, because LOG(FATAL) throws an exception.

python/dlr/dlr_model.py
```diff
@@ -120,15 +120,19 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0):
...
+        self.output_dtypes = []
+        
         for i in range(self.num_weights):
             self.weight_names.append(self._get_weight_name(i))
 
         self.num_outputs = self._get_num_outputs()
         self._lazy_init_output_shape()
+        self._fetch_input_names()
+        self._fetch_output_names()
+        self._fetch_output_dtypes()
```
apivovarov, line 135: Lets add `self._fetch_input_dtypes()` too

python/dlr/dlr_model.py
apivovarov: `if self.has_metadata()` should be removed

python/dlr/dlr_model.py
```diff
@@ -160,15 +164,42 @@ def get_input_names(self):
...
+                dtype = c_char_p()
+                _check_call(_LIB.GetDLROutputType(byref(self.handle), i, byref(dtype)))
+                self.output_dtypes.append(dtype.value.decode('utf-8'))
+
     def get_output_names(self):
-        """
-        Get all output names
+        if not self.has_metadata():
+            raise DLRError("Model has no metadata!")
+        else:
```
apivovarov, line 195: `else` can be removed

python/dlr/dlr_model.py
```diff
@@ -160,15 +164,42 @@ def get_input_names(self):
...
+            return self.output_names
 
-        Returns
-        -------
-        out : list of :py:class:`str`
-        """
-        raise NotImplementedError
+    def get_output_dtypes(self):
+        if not self.has_metadata():
```
apivovarov, line 199: Do not check for metadata here. Just `return self.output_dtypes`

src/dlr.cc
```diff
@@ -123,6 +123,38 @@ extern "C" int GetDLRNumOutputs(DLRModelHandle* handle, int* num_outputs) {
...
+
+extern "C" int GetDLROutputName(DLRModelHandle* handle, const int index, const char** name) {
+  API_BEGIN();
+  DLRModel* model = static_cast<DLRModel*>(*handle);
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  *name = model->GetOutputName(index);
+  API_END();
+}
+
+extern "C" int GetDLROutputIndex(DLRModelHandle* handle, char* name, int* index) {
```
apivovarov, line 142: `const char* name`

src/dlr.cc
apivovarov: `const char* name`

src/dlr_common.cc
apivovarov: Return `#include "dlr_common.h"` on top.

src/dlr_tvm.cc
```diff
@@ -214,3 +223,32 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+const char* TVMModel::GetOutputName(const int index) const {
+  std::string name_str = this->metadata["Model"]["Outputs"][index]["name"];
+  return name_str.c_str();
+}
+
+int TVMModel::GetOutputIndex(const char* name) const {
+  int idx = -1;
+
+  for (int i = 0; i < this->num_outputs_; i++) {
+    std::string name_str = this->metadata["Model"]["Outputs"][i]["name"];
```
apivovarov, line 240: Lets add `vector<string> output_names_` to `TVMModel` and fill it with output names from metadata in the constructor. Then we can use `output_names_` vector instead of using `metadata` each time users call `GetOutputIndex`.

src/dlr_tvm.cc
```diff
@@ -214,3 +223,32 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+    if (strcmp(name, name_str.c_str()) == 0) {
+      idx = i;
+      break;
```
apivovarov, line 243: just  `return i`

src/dlr_tvm.cc
```diff
@@ -214,3 +223,32 @@ void TVMModel::UseCPUAffinity(bool use) {
     LOG(INFO) << "CPU Affinity is disabled";
   }
 }
+
+bool TVMModel::HasMetadata() const {
+  return this->metadata != nullptr;
+}
+
+const char* TVMModel::GetOutputName(const int index) const {
+  std::string name_str = this->metadata["Model"]["Outputs"][index]["name"];
```
apivovarov, line 232: Lets use `vector<string> output_names_` which is filled with output names from `metadata` in the constructor.

include/dlr_tvm.h
cnv1989: I don't think that is a mandate. It depends how we pull the file and what directory use in includes.

I can get rid of json.hpp here.

include/dlr_common.h
cnv1989: ok.

src/dlr_tvm.cc
cnv1989: There is no benefit of storing it in two places. There is no drop in efficiency from the looks of it.

src/dlr_tvm.cc
cnv1989: This will cause issues i weird cases where the output names won't match or number of outputs come out to be zero. Also if uses passes bad output name it will always return the last index which is incorrect.

src/dlr.cc
cnv1989: 👍

python/dlr/dlr_model.py
cnv1989: Yes It can be removed 👍

python/dlr/dlr_model.py
cnv1989: 👍

include/dlr_common.h
cnv1989: This is a virtual function `// unreachable ` is misleading.

include/dlr.h
cnv1989: 👍

src/dlr_tvm.cc
trevor-m: We can do
```
for (...) {
  if (...) {
    return i;
  }
}
return -1;
```

python/dlr/api.py
cnv1989: yes 👍

demo/cpp/model_peeker.cc
cnv1989: 👍

src/dlr_tvm.cc
cnv1989: Same thing differently. Ok I will do it if it makes you happy ;)

src/dlr_tvm.cc
cnv1989: I don't think its necessary. Why do we need to do that?

cnv1989: Adding unecessary indirection to the actual data.

trevor-m: I agree we shouldn't need to create `output_names_`.

However this function should return
```
return this->metadata["Model"]["Outputs"][index]["name"].c_str();
```

Since `name_str.c_str()` will be invalid when `name_str` goes out of scope (when this function ends).

src/dlr_common.cc
```diff
@@ -1,7 +1,7 @@
-#include "dlr_common.h"
-
 #include <dmlc/filesystem.h>
-
+#include <iostream>
```
apivovarov, line 2: Looks like we do not need it

python/dlr/dlr_model.py
```diff
@@ -120,15 +120,21 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0):
...
+        
         for i in range(self.num_weights):
             self.weight_names.append(self._get_weight_name(i))
 
         self.num_outputs = self._get_num_outputs()
         self._lazy_init_output_shape()
+        self._fetch_input_names()
+        self._fetch_output_names()
+        self._fetch_input_dtypes()
+        self._fetch_output_dtypes()
```
apivovarov, line 137: `_fetch_output_dtypes` and `_fetch_input_dtypes` will work for TVM, Treelite and TFLite backends. Other backends do not support it yet.
So, we should have try catch here as well

src/dlr_tvm.cc
```diff
@@ -214,3 +223,29 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+  return name_str.c_str();
+}
+
+int TVMModel::GetOutputIndex(const char* name) const {
+  for (int i = 0; i < this->num_outputs_; i++) {
+    std::string name_str = this->metadata["Model"]["Outputs"][i]["name"];
+    if (strcmp(name, name_str.c_str()) == 0) {
```
trevor-m, line 239: nit: use `name_str == name`

python/dlr/dlr_model.py
```diff
@@ -160,15 +166,41 @@ def get_input_names(self):
...
-        """
-        Get all output names
+    def has_metadata(self) -> bool:
+        flag = ctypes.c_bool()
+        _check_call(_LIB.GetDLRHasMetadata(byref(self.handle), byref(flag)))
+        return flag.value
+    
+    def _fetch_output_names(self):
+        self.output_names = []
+        if self.has_metadata():
```
apivovarov, line 176: Some backends/models might give you output names without metadata file (e.g. Tensorflow). We should try to fetch `output_names` and if `GetDLROutputName` throws exception we should leave self.output_names empty.

python/dlr/dlr_model.py
cnv1989: 👍

src/dlr_tvm.cc
cnv1989: 👍

src/dlr_tvm.cc
cnv1989: 👍

src/dlr_tvm.cc
```diff
@@ -214,3 +223,34 @@ void TVMModel::UseCPUAffinity(bool use) {
...
+    }
+  } 
+  return -1;
+}
+
+void TVMModel::GetOuputByName(const char* name, float* out) {
+  int output_index = this->GetOutputIndex(name);
+  if (output_index >= 0) {
+    this->GetOutput(output_index, out);
+  }
```
apivovarov, line 255: It will silently do nothing if name is not found

python/dlr/dlr_model.py
```diff
@@ -160,15 +166,91 @@ def get_input_names(self):
...
+            raise Exception(f"Index cannot be greater than {self.num_outputs - 1}")
+        return self.get_ouput_names()[index]
+
+    def get_input_dtype(self, index):
+        if index < 0 or index >= self.num_inputs:
+            raise Exception(f"Index cannot be greater than {self.num_inputs - 1}")
+        return self.get_input_dtypes()[index]
+
+    def get_output_dtype(self, index):
+        if index < 0 or index >= self.num_outputs:
```
apivovarov, line 251: `not(0 <= x < self.num_outputs)`

python/dlr/api.py
```diff
@@ -118,3 +118,53 @@ def get_version(self):
...
+            return self._impl.get_input_dtype(index)
+        except Exception as ex:
+            self.neo_logger.exception("error in getting input data type {} {}".format(self._impl.__class__.__name__, ex))
+            raise ex
+
+    def get_output_dtype(self, index):
+        try:
+            return self._impl.get_output_dtype(index)
+        except Exception as ex:
+            self.neo_logger.exception("error in getting input data type {} {}".format(self._impl.__class__.__name__, ex))
```
apivovarov, line 168: output data type, not input

cnv1989: 👍

PR #174, opened 2020-05-15
Update TVM to neo-ai/dev (0.7dev1), Bump to DLR 1.2

trevor-m:
This PR will update TVM to match neo-ai/dev, which is on TVM version 0.7dev1.

This will bring in new features such as heterogeneous TRT support which involves a new format for TRT-enabled models. Neo compiled models which used the TensorRT integration that ran using DLR 1.0 or 1.1 will be upgraded to this new format automatically using the compatibility.py scripts in this PR.

Also included are changes to the CI to accommodate new requirements of TVM including Cmake 3.13, C++14. This is accomplished by using a Dockerfile for the GPU build which will make it easier to update in the future as well - previously we had some AMI with manually installed dependencies.


CMakeLists.txt
```diff
@@ -207,60 +220,66 @@ if(USE_TENSORRT)
...
+         ${TVM_SRC}/src/ir/attrs.cc
 
          ${TVM_SRC}/src/relay/ir/expr_functor.cc
          ${TVM_SRC}/src/relay/ir/pattern_functor.cc
-         ${TVM_SRC}/src/relay/pass/util.cc
-         ${TVM_SRC}/src/relay/ir/type_functor.cc
-         ${TVM_SRC}/src/relay/ir/type.cc
-         ${TVM_SRC}/src/lang/ir.cc
-         ${TVM_SRC}/src/lang/expr.cc
+         ${TVM_SRC}/src/relay/op/type_relations.cc
```
zhiics, line 228: Can we use GLOB here to get the files? It looks not very safe to add them one by one IMH.

python/dlr/compatibility.py
```diff
@@ -0,0 +1,208 @@
...
+    to_ver : str
+        The target version.
+
+    Returns
+    -------
+    fupdater : function
+        The updater function
+    """
+    def _updater(data):
+        # assert data["attrs"]["tvm_version"].startswith(from_ver)
```
zhiics, line 43: enable or remove

python/dlr/compatibility.py
```diff
@@ -0,0 +1,208 @@
...
+        "relay.ModulePass": _rename("transform.ModulePass"),
+        "relay.Sequential": _rename("transform.Sequential"),
+        "relay.attrs.ResizeAttrs": _update_resize_attrs,
+        # TIR
+        "Variable": _update_tir_var("tir.Var"),
+        "SizeVar": _update_tir_var("tir.SizeVar"),
+    }
+    return create_updater(node_map, "0.6", "0.7")
+
+def check_tensorrt_compatibility(file_path):
```
zhiics, line 138: I think this is a temporary solution it seems not quite safe to just update the strings as they may change again and there might be more nodes. For the long term, we should probably retire the old version or ask for update to the newer version.

tests/ci_build/ci_build.sh
```diff
@@ -62,11 +62,11 @@ if [ "$#" -lt 1 ] || [ ! -e "${SCRIPT_DIR}/Dockerfile.${CONTAINER_TYPE}" ]; then
 fi
 
 # Use nvidia-docker if the container is GPU.
-if [[ "${CONTAINER_TYPE}" == *"gpu"* ]]; then
-    DOCKER_BINARY="nvidia-docker"
-else
-    DOCKER_BINARY="docker"
-fi
+#if [[ "${CONTAINER_TYPE}" == *"gpu"* ]]; then
```
zhiics, line 65: remove? just keep docker?

CMakeLists.txt
trevor-m: I would prefer to use glob as well. I did that originally but there was some concern about increasing the size of the binary. I think the size was around 7mb with glob and ~3mb with this approach. It is quite a painstaking process to pick out the files one by one.

python/dlr/compatibility.py
trevor-m: Hi Zhi, I agree. This is a temporarily solution to allow some degree of backwards compatibility until we can include DLR runtime with the compiled artifacts.

CMakeLists.txt
zhiics: NVM, we can keep it for now. We will need to work on the JSON runtime later. Once that it ready, we should be able remove most of them (probably even all) as only runtime data structures and files are needed.

tests/ci_build/Dockerfile.cpu_bare
```diff
@@ -8,19 +8,20 @@ ENV DEBIAN_FRONTEND noninteractive
...
+    yum install -y centos-release-scl && \
+    yum install -y devtoolset-7 && \
+    yum install -y devtoolset-7-gcc devtoolset-7-binutils devtoolset-7-gcc-c++ && \
     # Python
     wget https://repo.continuum.io/miniconda/Miniconda2-4.3.27-Linux-x86_64.sh && \
     bash Miniconda2-4.3.27-Linux-x86_64.sh -b -p /opt/python && \
     # CMake
-    wget -nv -nc https://cmake.org/files/v3.6/cmake-3.6.0-Linux-x86_64.sh --no-check-certificate && \
-    bash cmake-3.6.0-Linux-x86_64.sh --skip-license --prefix=/usr
+    wget -nv -nc https://cmake.org/files/v3.13/cmake-3.13.0-Linux-x86_64.sh --no-check-certificate && \
```
apivovarov, line 18: If possible lets use the latest cmake `3.17.2` everywhere or at least `3.14.0` (which is used in `container/Dockerfile.gpu`)

PR #186, opened 2020-05-29
Check for metadata keys instead of try/catch in GetOutputName

trevor-m:
Attempting to access the metadata output names when they weren't in the metadata file was causing segfaults. This could be due to:
1. Behavior of map to initialize new objects when `[]` operator is used instead of throwing `nlohmann::detail::type_error`, then calling `.get_ref<const std::string&>().c_str()` on these empty objects.
2. Possible due to not handling all of the different exception types, for example `nlohmann::detail::out_of_range_error`. 

To make things simpler and reduce chance of not catching unknown exception, we instead explicitly check if the keys exist.

### DIFF
--- a/src/dlr_tvm.cc
+++ b/src/dlr_tvm.cc
@@ -231,14 +231,15 @@ const char* TVMModel::GetOutputName(const int index) const {
     LOG(INFO) << "No metadata file was found!";
     return nullptr;
   }
-  try {
+  if (this->metadata.count("Model") &&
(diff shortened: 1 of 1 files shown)


apivovarov:
BTW, I noticed that we use `metadata == nullptr` in our code. It is not obvious what exactly overloaded '==' op is doing with `nullptr` and why freshly created object `metadata` returns true for that.

PR #188, opened 2020-06-01
[Windows] Fix compilation errors due to misinterpretation of TVM_DLL on Windows

ymwangg:
This PR removed the unused .cc files that cause compilation errors on Windows. The functions in these removed .cc files are already in the tvm_runtime_static.lib file and do no need to be explicitly requested here. Please merge this after https://github.com/neo-ai/tvm/pull/110 is merged first.


trevor-m:
We need sort.cc for models which use that external function. I think the object detections modes use that.

zhiics:
Should we enable sort through config.cmake instead of adding it to the DLR cmakefile?

trevor-m:
> Should we enable sort through config.cmake instead of adding it to the DLR cmakefile?

Possibly, but then everyone will need to specify USE_SORT when building DLR if they want to run object detection models. Most DLR users will not know whether their model needs USE_SORT so this would be not a good experience.

Perhaps we can exclude sort.cc for windows only.

zhiics:
yeah, but this would also enable sort if even users don't want, right? It's just where they want to enable it. We could probably add a switch in this file to disable it. But this is probably not platform specific, but model specific. Disabling it for windows would also cause errors on OD models.

ymwangg:
I have checked only `${TVM_SRC}/src/runtime/cpu_device_api.cc` causes the compilation failure.

trevor-m:
If a user compiles a model with Neo, they have no idea whether it used sort or not for their model. There isn't really any negative to always include sort.  

Disabling it for windows would not cause any more issues because that file cannot compile at all for windows at the moment. @ymwangg can you confirm if sort.cc cannot compile for windows and what the error is?

ymwangg:
> If a user compiles a model with Neo, they have no idea whether it used sort or not for their model. There isn't really any negative to always include sort.
> 
> Disabling it for windows would not cause any more issues because that file cannot compile at all for windows at the moment. @ymwangg can you confirm if sort.cc cannot compile for windows and what the error is?

zhiics:
Yeah, my point is if we can also always enable it, we can probably just enable in the TVM side as it is not part of DLR. But your point of that users may need to be aware of that makes sense.

ymwangg:
> If a user compiles a model with Neo, they have no idea whether it used sort or not for their model. There isn't really any negative to always include sort.
> 
> Disabling it for windows would not cause any more issues because that file cannot compile at all for windows at the moment. @ymwangg can you confirm if sort.cc cannot compile for windows and what the error is?

I have confirmed only `${TVM_SRC}/src/runtime/cpu_device_api.cc` needs to be removed to fix the compilation error.

zhiics:
okay, since removing cpu_device_api.cc solves the problem, we can just make the change minimal.

.gitmodules
```diff
@@ -1,7 +1,6 @@
 [submodule "3rdparty/tvm"]
 	path = 3rdparty/tvm
 	url = https://github.com/neo-ai/tvm.git
-        branch = stable
```
zhiics, line 4: do we need to change this?

ymwangg: this change is made by https://github.com/neo-ai/neo-ai-dlr/pull/189 that has been merged in the master branch.

ymwangg: > Shouldn't we revert the changes from tvm?

The changes from tvm are already up to date with master.

zhiics: I see. You need to rebase your change again the master branch of dlr because you are currently incorporating other commits.

ymwangg: rebased.

CMakeLists.txt
```diff
@@ -97,9 +97,8 @@ include_directories("${JSON_SRC}")
 # Add only top level *.cc files (non-RECURSE)
 FILE(GLOB DLR_SRC
     "src/*.cc"
-    ${TVM_SRC}/src/runtime/dso_library.cc
-    ${TVM_SRC}/src/runtime/cpu_device_api.cc
-    ${TVM_SRC}/src/runtime/contrib/sort/sort.cc
+     ${TVM_SRC}/src/runtime/dso_library.cc
+     ${TVM_SRC}/src/runtime/contrib/sort/sort.cc
```
apivovarov, line 101: why 5 spaces?

ymwangg:
After some investigation, I found cherry-pick tvm runtime .cc files added to `DLR_SRC` is necessary. Though remove those files won't cause compilation error, but it will cause runtime errors. It seems that some functions in these .cc files are not defined in the tvm runtime .h files and cannot be found as dependency by cmake. Instead, we manually define the `TVM_EXPORT` compilation flag for the compilation of these .cc files.

PR #197, opened 2020-06-09
Use dlr binary from compilation artifact if avaliable

cnv1989:
# Change Description

Going forward neo will be shipping dlr binary along with model artifact. If the model artifact contains the binary we want the dlr to use that over the default binary that is downloaded as part of pip install. This PR updates the DLRModelImpl to load the binary from model artifact if available.

# Testing

## Force Default
```python
from dlr import DLRModel
import numpy as np
```

```python
from pathlib import Path
```

```python
compiled = Path("resnet_v1.5_50-ml_c4").resolve()
```

```python
model = DLRModel(compiled.as_posix(), use_default_dlr=True)
```

    2020-06-10 23:14:49,086 INFO Using default libdlr.so from /home/DLR/neo-ai-dlr/python/dlr/../../build/lib/libdlr.so

```python
model.get_input_name(0)
```

    'input_tensor'

```python
model.get_output_name(1)
```

    'softmax_tensor:0'

```python
model.get_output_name(1)
```

    'softmax_tensor:0'

```python
model.has_metadata()
```

    True

```python
# model.get_weight_names()
```

```python
dog_image_path = Path("/home/DLR/neo-ai-dlr/tests/python/integration/dog.npy")
```

```python
dog = np.load(dog_image_path).astype(np.float32)
```

```python
dogT = dog.transpose([0, 2, 3, 1])
```

```python
dogT.shape
```

    (1, 224, 224, 3)

```python
input_name = "input_tensor"
```

```python
input_dim = dogT.ndim
```

```python
model.run({input_name: dogT})
```

    [array([447], dtype=int32),
     array([[5.5828092e-05, 1.5310821e-04, 8.4513071e-04, ..., 1.2091993e-04,
             8.1732614e-05, 1.8214799e-03]], dtype=float32)]

```python

```

## Default Fallback
Fallback to default libdlr when there is no libdlr.so in the model artifact

```python
from dlr import DLRModel
import numpy as np
```

```python
from pathlib import Path
```

```python
compiled = Path("resnet_v1.5_50-ml_c4").resolve()
```

```python
model = DLRModel(compiled.as_posix())
```

    2020-06-10 23:17:38,724 INFO Using default libdlr.so from /home/DLR/neo-ai-dlr/python/dlr/../../build/lib/libdlr.so

```python
model.get_input_name(0)
```

    'input_tensor'

```python
model.get_output_name(1)
```

    'softmax_tensor:0'

```python
model.get_output_name(1)
```

    'softmax_tensor:0'

```python
model.has_metadata()
```

    True

```python
# model.get_weight_names()
```

```python
dog_image_path = Path("/home/DLR/neo-ai-dlr/tests/python/integration/dog.npy")
```

```python
dog = np.load(dog_image_path).astype(np.float32)
```

```python
dogT = dog.transpose([0, 2, 3, 1])
```

```python
dogT.shape
```

    (1, 224, 224, 3)

```python
input_name = "input_tensor"
```

```python
input_dim = dogT.ndim
```

```python
model.run({input_name: dogT})
```

    [array([447], dtype=int32),
     array([[5.5828092e-05, 1.5310821e-04, 8.4513071e-04, ..., 1.2091993e-04,
             8.1732614e-05, 1.8214799e-03]], dtype=float32)]

```python

```

## Binary from Model Artifact

```python
from dlr import DLRModel
import numpy as np
```

```python
from pathlib import Path
```

```python
compiled = Path("resnet_v1.5_50-ml_c4").resolve()
```

```python
model = DLRModel(compiled.as_posix())
```

    2020-06-10 23:19:49,443 INFO Found libdlr.so in model artifact. Using /home/DLR/resnet_v1.5_50-ml_c4/libdlr.so

```python
model.get_input_name(0)
```

    'input_tensor'

```python
model.get_output_name(1)
```

    'softmax_tensor:0'

```python
model.get_output_name(1)
```

    'softmax_tensor:0'

```python
model.has_metadata()
```

    True

```python
# model.get_weight_names()
```

```python
dog_image_path = Path("/home/DLR/neo-ai-dlr/tests/python/integration/dog.npy")
```

```python
dog = np.load(dog_image_path).astype(np.float32)
```

```python
dogT = dog.transpose([0, 2, 3, 1])
```

```python
dogT.shape
```

    (1, 224, 224, 3)

```python
input_name = "input_tensor"
```

```python
input_dim = dogT.ndim
```

```python
model.run({input_name: dogT})
```

    [array([447], dtype=int32),
     array([[5.5828092e-05, 1.5310821e-04, 8.4513071e-04, ..., 1.2091993e-04,
             8.1732614e-05, 1.8214799e-03]], dtype=float32)]

```python

```

### DIFF
--- a/python/dlr/api.py
+++ b/python/dlr/api.py
@@ -79,7 +79,7 @@ class DLRModel(IDLRModel):
                 dev_type = 'cpu'
             if dev_id is None:
                 dev_id = 0
-            self._impl = DLRModelImpl(model_path, dev_type, dev_id, use_default_dlr, self.neo_logger)
+            self._impl = DLRModelImpl(model_path, dev_type, dev_id, error_log_file, use_default_dlr)
         except Exception as ex:
             self.neo_logger.exception("error in DLRModel instantiation {}".format(ex))
             raise ex
--- a/python/dlr/api.py
+++ b/python/dlr/api.py
@@ -50,7 +50,7 @@ def _is_module_found(name):
 
 # Wrapper class
 class DLRModel(IDLRModel):
-    def __init__(self, model_path, dev_type=None, dev_id=None, error_log_file=None):
+    def __init__(self, model_path, dev_type=None, dev_id=None, error_log_file=None, use_default_dlr=False):
         self.neo_logger = create_logger(log_file=error_log_file)
         try:
             # Find correct runtime implementation for the model
@@ -79,7 +79,7 @@ class DLRModel(IDLRModel):
                 dev_type = 'cpu'
             if dev_id is None:
                 dev_id = 0
-            self._impl = DLRModelImpl(model_path, dev_type, dev_id)
+            self._impl = DLRModelImpl(model_path, dev_type, dev_id, use_default_dlr, self.neo_logger)
         except Exception as ex:
             self.neo_logger.exception("error in DLRModel instantiation {}".format(ex))
             raise ex
--- a/python/dlr/dlr_model.py
+++ b/python/dlr/dlr_model.py
@@ -3,11 +3,12 @@ import ctypes
 from ctypes import c_void_p, c_int, c_char_p, byref, POINTER, c_longlong
 import numpy as np
 import os
+import sys
 from .api import IDLRModel
 from .compatibility import check_tensorrt_compatibility
-
 from .libpath import find_lib_path
 
+<<<<<<< HEAD
 # Map from dtype string to ctype type.
 # Equivalent to np.ctypeslib.as_ctypes_type which requires numpy>=1.16.1
 DTYPE_TO_CTYPE = {
@@ -33,28 +34,15 @@ def _get_ctype_from_dtype(dtype):
     if dtype not in DTYPE_TO_CTYPE:
         raise ValueError("Model has input or output datatype {} which is not supported.".format(dtype))
     return DTYPE_TO_CTYPE[dtype]
+=======
+>>>>>>> Use dlr binary from compilation artifact if avaliable
 
 class DLRError(Exception):
     """Error thrown by DLR"""
     pass
 
-def _check_call(ret):
-    """
-    Check the return value of C API call
-    This function will raise exception when error occurs.
-    Wrap every API call with this function
-
-    Parameters
-    ----------
-    ret : int
-        return value from API calls
-    """
-    if ret != 0:
-        raise DLRError(_LIB.DLRGetLastError().decode('ascii'))
-
-def _load_lib():
+def _load_lib(lib_paths):
     """Load DLR library."""
-    lib_paths = find_lib_path()
     if len(lib_paths) == 0:
         return None
     try:
@@ -87,9 +75,6 @@ def _load_lib():
     lib.DLRGetLastError.restype = ctypes.c_char_p
     return lib
 
-# load the DLR library globally
-_LIB = _load_lib()
-
 class DLRModelImpl(IDLRModel):
     """
     Load a Neo-compiled model
@@ -113,16 +98,47 @@ class DLRModelImpl(IDLRModel):
 
     def _parse_backend(self):
         backend = c_char_p()
-        _check_call(_LIB.GetDLRBackend(byref(self.handle),
+        self._check_call(self._lib.GetDLRBackend(byref(self.handle),
                                        byref(backend)))
         return backend.value.decode('ascii')
 
     def _get_version(self):
         version = c_char_p()
-        _check_call(_LIB.GetDLRVersion(byref(version)))
+        self._check_call(self._lib.GetDLRVersion(byref(version)))
         return version.value.decode('ascii')
 
-    def __init__(self, model_path, dev_type='cpu', dev_id=0):
+    def _check_call(self, ret):
+        """
+        Check the return value of C API call
+        This function will raise exception when error occurs.
+        Wrap every API call with this function
+
+        Parameters
+        ----------
+        ret : int
+            return value from API calls
+        """
+        if ret != 0:
+            raise DLRError(self._lib.DLRGetLastError().decode('ascii'))
+
+    def _init_libdlr(self):
+        paths = find_lib_path()
+        
+        local_libdlr_so_path = os.path.join(self.model_path, "libdlr.so")
+        if sys.platform == 'win32':
+            local_libdlr_so_path = os.path.join(self.model_path, "dlr.dll")
+        elif sys.platform == 'darwin':
+            local_libdlr_so_path = os.path.join(self.model_path, "dlr.dylib")
+        
+        if not self.use_default_dlr and os.path.exists(local_libdlr_so_path):
+            self.logger.info(f"Found libdlr.so in model artifact. Using {local_libdlr_so_path}")
+            paths = [local_libdlr_so_path]
+        else:
+            self.logger.info(f"Using default libdlr.so from {paths[0]}")
+        self._lib = _load_lib(paths)
+
+    def __init__(self, model_path, dev_type='cpu', dev_id=0, use_default_dlr=False, logger=None):
+        self.logger = logger
         if not os.path.exists(model_path):
             raise ValueError("model_path %s doesn't exist" % model_path)
         # Backwards compatibility for .tensorrt artifacts.
@@ -135,7 +151,10 @@ class DLRModelImpl(IDLRModel):
             'gpu': 2,
             'opencl': 4,
         }
-        _check_call(_LIB.CreateDLRModel(byref(self.handle),
+        self.model_path = model_path
+        self.use_default_dlr = use_default_dlr
+        self._init_libdlr()
+        self._check_call(self._lib.CreateDLRModel(byref(self.handle),
                                         c_char_p(model_path.encode()),
                                         c_int(device_table[dev_type]),
                                         c_int(dev_id)))
@@ -166,20 +185,20 @@ class DLRModelImpl(IDLRModel):
     def __del__(self):
         if getattr(self, "handle", None) is not None and self.handle is not None:
             if getattr(self, "lib", None) is not None:
-                _check_call(_LIB.DeleteDLRModel(byref(self.handle)))
+                self._check_call(self._lib.DeleteDLRModel(byref(self.handle)))
             self.handle = None
 
     def _get_num_inputs(self):
         """Get the number of inputs of a network"""
         num_inputs = c_int()
-        _check_call(_LIB.GetDLRNumInputs(byref(self.handle),
+        self._check_call(self._lib.GetDLRNumInputs(byref(self.handle),
                                          byref(num_inputs)))
         return num_inputs.value
 
     def _get_num_weights(self):
         """Get the number of weights of a network"""
         num_weights = c_int()
-        _check_call(_LIB.GetDLRNumWeights(byref(self.handle),
+        self._check_call(self._lib.GetDLRNumWeights(byref(self.handle),
                                           byref(num_weights)))
         return num_weights.value
 
@@ -195,7 +214,7 @@ class DLRModelImpl(IDLRModel):
 
     def has_metadata(self) -> bool:
         flag = ctypes.c_bool()
-        _check_call(_LIB.GetDLRHasMetadata(byref(self.handle), byref(flag)))
+        self._check_call(self._lib.GetDLRHasMetadata(byref(self.handle), byref(flag)))
         return flag.value
     
     def _fetch_output_names(self):
@@ -203,7 +222,7 @@ class DLRModelImpl(IDLRModel):
         try:
             for i in range(self.num_outputs):
                 name = c_char_p()
-                _check_call(_LIB.GetDLROutputName(byref(self.handle), i, byref(name)))
+                self._check_call(self._lib.GetDLROutputName(byref(self.handle), i, byref(name)))
                 self.output_names.append(name.value.decode('utf-8'))
         except Exception:
             """
@@ -223,7 +242,7 @@ class DLRModelImpl(IDLRModel):
         try:
             for i in range(self.num_inputs):
                 dtype = c_char_p()
-                _check_call(_LIB.GetDLRInputType(byref(self.handle), i, byref(dtype)))
+                self._check_call(self._lib.GetDLRInputType(byref(self.handle), i, byref(dtype)))
                 self.input_dtypes.append(dtype.value.decode('utf-8'))
         except Exception:
             """
@@ -237,7 +256,7 @@ class DLRModelImpl(IDLRModel):
         try:
             for i in range(self.num_outputs):
                 dtype = c_char_p()
-                _check_call(_LIB.GetDLROutputType(byref(self.handle), i, byref(dtype)))
+                self._check_call(self._lib.GetDLROutputType(byref(self.handle), i, byref(dtype)))
                 self.output_dtypes.append(dtype.value.decode('utf-8'))
         except Exception:
             """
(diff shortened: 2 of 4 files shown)


python/dlr/dlr_model.py
```diff
@@ -87,16 +70,47 @@ def _lazy_init_output_shape(self):
...
+        paths = find_lib_path()
+        
+        local_libdlr_so_path = os.path.join(self.model_path, "libdlr.so")
+        if sys.platform == 'win32':
+            local_libdlr_so_path = os.path.join(self.model_path, "dlr.dll")
+        elif sys.platform == 'darwin':
+            local_libdlr_so_path = os.path.join(self.model_path, "dlr.dylib")
+        
+        if not self.use_default_dlr and os.path.exists(local_libdlr_so_path):
+            self.logger.info(f"Found libdlr.so in model artifact. Using {local_libdlr_so_path}")
```
trevor-m, line 106: Lets avoid f-string formatting because we have to support python 3.5 for some devices.

cnv1989: Good catch 👍

python/dlr/api.py
```diff
@@ -50,7 +50,7 @@ def _is_module_found(name):
 
 # Wrapper class
 class DLRModel(IDLRModel):
-    def __init__(self, model_path, dev_type=None, dev_id=None, error_log_file=None):
+    def __init__(self, model_path, dev_type=None, dev_id=None, error_log_file=None, use_default_dlr=False):
```
apivovarov, line 53:
```
class DLRModelImpl(IDLRModel):
    def __init__(self, model_path, dev_type='cpu', dev_id=0, use_default_dlr=False, logger=None):
```

cnv1989: logger is not same as error_log_file

python/dlr/dlr_model.py
```diff
@@ -87,16 +70,47 @@ def _lazy_init_output_shape(self):
...
+        Parameters
+        ----------
+        ret : int
+            return value from API calls
+        """
+        if ret != 0:
+            raise DLRError(self._lib.DLRGetLastError().decode('ascii'))
+
+    def _init_libdlr(self):
+        paths = find_lib_path()
```
apivovarov, line 97: Can we move it to line 109?

python/dlr/dlr_model.py
```diff
@@ -87,16 +70,47 @@ def _lazy_init_output_shape(self):
...
+        
+        local_libdlr_so_path = os.path.join(self.model_path, "libdlr.so")
+        if sys.platform == 'win32':
+            local_libdlr_so_path = os.path.join(self.model_path, "dlr.dll")
+        elif sys.platform == 'darwin':
+            local_libdlr_so_path = os.path.join(self.model_path, "dlr.dylib")
```
apivovarov, line 103: no need to repeat `local_libdlr_so_path = os.path.join(self.model_path,` 3 times. Use `if` or `dict` to determine `libFileName` and call `os.path.join` once.

python/dlr/api.py
apivovarov: Lets use the same name for parameters and make sure that `error_log_file` is before `use_default_dlr`

cnv1989: Do we want to pass the error_log_file, Why instantiate the logger multiple times? Its better to pass the logger with the right name we should avoid confusing names. I can fix the order

python/dlr/dlr_model.py
cnv1989: Could you elaborate?

python/dlr/api.py
cnv1989: Looking at it again, it is better to keep it consistent. 👍

python/dlr/dlr_model.py
```diff
@@ -157,20 +112,68 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0, use_default_dlr=False,
...
+        ----------
+        ret : int
+            return value from API calls
+        """
+        if ret != 0:
+            raise DLRError(DLRModelImpl._LIB.DLRGetLastError().decode('ascii'))
+
+    def _init_libdlr(self):
+        if DLRModelImpl._LIB is None:
+            lib_path = Path(find_lib_path())
```
trevor-m, line 152: Change name to `system_lib_path` to differentiate more from`local_dlr_path`

cnv1989: 👍

python/dlr/dlr_model.py
apivovarov: calling `find_lib_path()` is needed only If `not self.use_default_dlr and os.path.exists(local_libdlr_so_path)` is false. We can move `find_lib_path()` to `else` block - line 109

python/dlr/dlr_model.py
```diff
@@ -140,20 +112,69 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0):
...
+
+    def _init_libdlr(self):
+        if DLRModelImpl._LIB is None:
+            system_lib_path = Path(find_lib_path())
+            local_dlr_path = Path(self.model_path).joinpath(system_lib_path.name)
+
+            if self.use_default_dlr:
+                self.logger.info("Forced to use default {} from {}".format(system_lib_path.name, system_lib_path.as_posix()))
+                DLRModelImpl._LIB = _load_lib(system_lib_path)
+            elif local_dlr_path.exists():
```
apivovarov, line 158: We do not need to call `find_lib_path()` in case `local_dlr_path.exists()`

python/dlr/dlr_model.py
```diff
@@ -140,20 +112,69 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0):
...
+        Parameters
+        ----------
+        ret : int
+            return value from API calls
+        """
+        if ret != 0:
+            raise DLRError(DLRModelImpl._LIB.DLRGetLastError().decode('ascii'))
+
+    def _init_libdlr(self):
+        if DLRModelImpl._LIB is None:
```
apivovarov, line 151: if DLRModelImpl._LIB is not None:
    log.info("already loaded") and return
the rest of the code can be shifted to the left

python/dlr/libpath.py
```diff
@@ -48,4 +51,6 @@ def find_lib_path():
         raise DLRLibraryNotFound(
             'Cannot find DLR Library in the candidate path, ' +
             'List of candidates:\n' + ('\n'.join(dll_path)))
-    return lib_path
+
+    # If multiple paths are found always prefer the one that is in system paths.
+    return lib_path[0]
```
apivovarov, line 56: We do not need to check all `dll_path` for existence (on line 50) `lib_path = [...]

python/dlr/dlr_model.py
```diff
@@ -140,20 +112,72 @@ def __init__(self, model_path, dev_type='cpu', dev_id=0):
...
+        if ret != 0:
+            raise DLRError(DLRModelImpl._LIB.DLRGetLastError().decode('ascii'))
+
+    def _init_libdlr(self):
+        if DLRModelImpl._LIB:
+            self.logger.info("{} is already loaded!".format(DLRModelImpl._LIB))
+            return
+
+        system_lib_path = Path(find_lib_path())
+        local_dlr_path = Path(self.model_path).joinpath(system_lib_path.name)
```
apivovarov, line 156: `find_lib_path()` will throw exception if DLR is not installed.

python/dlr/libpath.py
```diff
@@ -10,42 +10,53 @@ class DLRLibraryNotFound(Exception):
...
+            dll_paths.append(os.path.join(curr_path, './windows/Release/'))
+        libname = 'dlr.dll'
     elif sys.platform.startswith('linux') or sys.platform.startswith('freebsd'):
-        dll_path = [os.path.join(p, 'libdlr.so') for p in dll_path]
+        libname = 'libdlr.so'
     elif sys.platform == 'darwin':
-        dll_path = [os.path.join(p, 'libdlr.dylib') for p in dll_path]
+        libname = 'libdlr.dylib'
+
+    if not use_default_dlr and model_path and os.path.exists(os.path.join(model_path, libname)):
```
apivovarov, line 43: `and model_path is not None`

python/dlr/libpath.py
```diff
@@ -10,42 +10,53 @@ class DLRLibraryNotFound(Exception):
...
+        logger.info("Found {} in model artifact. Using dlr from {}".format(libname, os.path.join(model_path, libname)))
+        return  os.path.join(model_path, libname)
```
apivovarov, line 45: Format the file - extra spaces detected

python/dlr/dlr_model.py
```diff
@@ -77,26 +57,15 @@ class DLRModelImpl(IDLRModel):
     dev_id : int
         Device ID
     """
+    
+    # TVM crashes with `Check failed: override: Global PackedFunc runtime.module.loadfile_so is already registered`
+    # when we try to link muliple versions of the dlr withing the same process. In order to prevent this we make 
+    # _LIB a class attribute which is instantiated only once per class.
+    _LIB = None
```
apivovarov, line 64: Technically it does not matter where we keep `_LIB` on module level or inside the class as a Class Variable
Conceptually `_LIB` is module level attribute because it represent `libdlr.so` (we load single `libdlr.so` for the whole lifecycle of DLR module in Python process)
After that change we have to prefix `_LIB` with class name - `DLRModelImpl`.
There is no big reason to change `_LIB` to be Class Variable. 
Lets look at TVM as an example - in TVM `_LIB` is module level attribute - https://github.com/apache/incubator-tvm/blob/master/topi/python/topi/cpp/impl.py#L42

cnv1989: Module level instantiations (like this one https://github.com/apache/incubator-tvm/blob/master/topi/python/topi/cpp/impl.py#L42) is generally a bad practice. Any exception during the instantiation will cause the entire process to crash. Instead of keep then exception at instantiation this bubbles up the error to import time. This leads to non descriptive error messages and bad UX. I would say we should avoid patterns like the one in TVM as well.

cnv1989: More over to catch such crashes instead of doing global imports, we start doing local imports which causes memory leaks.

apivovarov: Technically it does not matter if we use short name `_LIB` or long name `DLRModelImpl._LIB`. What we are discussing here is variable naming question. Currently DLR and TVM use short name for `_LIB`. This PR puts `_LIB` to `DLRModelImpl` "namespace". As a result we have to use `DLRModelImpl._LIB` instead of short name `_LIB`. Moving `_LIB` to  `DLRModelImpl` "namespace" is not necessary for this particular PR. I'd open a separate PR for variable renaming / moving them to new namespaces.

python/dlr/dlr_model.py
```diff
@@ -3,66 +3,62 @@
...
-        finally:
-            os.environ['PATH'] = os.pathsep.join(pathBackup)
-    if not lib_success:
-        libname = os.path.basename(lib_paths[0])
+    
+    try:
+        # needed when the lib is linked with non-system-available dependencies
+        os.environ['PATH'] = os.pathsep.join(pathBackup + [os.path.dirname(lib_path)])
+        lib = ctypes.cdll.LoadLibrary(lib_path)
+        lib_success = True
```
apivovarov, line 30: is it unused var?

cnv1989: 👍

python/dlr/dlr_model.py
cnv1989: This is not a variable naming question. It's the context in which the variable is being instantiated and used. _LIB is a being instantiated in the class and used within the same class. 

Why _LIB is a class variable?

It depends on the model artifact's path (which is part of instance) to dynamically link the dlr.

trevor-m:
LGTM. Moving `_LIB` to `DLRModelImpl` makes sense from a design perspective because it shows that DLRModel is the one that owns the lib rather than the module.

@cnv1989 Please rebase and then we can merge.

apivovarov:
We still load `libdlr.so` only once during Python process lifecycle.
Before this PR we load `libdlr.so` when user calls `import dlr`.
Now we load `libdlr.so` when the first instance of `DLRModelImpl` is created.
What we change here is the event which triggers `libdlr.so` loading - it was `import dlr` before, and now it is  the first `DLRModelImpl` creation.
Conceptually nothing was changed. The relationship btw `libdlr.so` and DLR is the same - we still load single `libdlr.so` for the entire lyfecycle of the module.
`_LIB` is initialized by calling module level method `load_lib`. The event when we call this method was changed. But not the role and the ownership of LIB object.

python/dlr/dlr_model.py
apivovarov: Just to be clear of what change I'd like to see is to move `_LIB = None` from line 104 to line 86. The code will work the same as before.

python/dlr/dlr_model.py
```diff
@@ -103,26 +97,15 @@ class DLRModelImpl(IDLRModel):
     dev_id : int
         Device ID
     """
+    
+    # TVM crashes with `Check failed: override: Global PackedFunc runtime.module.loadfile_so is already registered`
+    # when we try to link muliple versions of the dlr withing the same process. In order to prevent this we make 
+    # _LIB a class attribute which is instantiated only once per class.
```
apivovarov, line 103: Lets replace this comment with the following:
```
# `libdlr.so` will be loaded when the first `DLRModelImpl` instance is created.
# Once `libdlr.so` is loaded it will be used to run the first and all consequent models.
```

python/dlr/dlr_model.py
```diff
@@ -3,11 +3,12 @@
 from ctypes import c_void_p, c_int, c_char_p, byref, POINTER, c_longlong
 import numpy as np
 import os
+import sys
 from .api import IDLRModel
 from .compatibility import check_tensorrt_compatibility
-
 from .libpath import find_lib_path
 
+<<<<<<< HEAD
```
cnv1989: Merge confict.

python/dlr/dlr_model.py
cnv1989: @apivovarov Lokesh informed me that this particular point has been clarified offline. Let me know if the PR is good to merge. Thanks!

lokeshgupta1975:
> We still load `libdlr.so` only once during Python process lifecycle.
> Before this PR we load `libdlr.so` when user calls `import dlr`.
> Now we load `libdlr.so` when the first instance of `DLRModelImpl` is created.
> What we change here is the event which triggers `libdlr.so` loading - it was `import dlr` before, and now it is the first `DLRModelImpl` creation.
> Conceptually nothing was changed. The relationship btw `libdlr.so` and DLR is the same - we still load single `libdlr.so` for the entire lyfecycle of the module.
> `_LIB` is initialized by calling module level method `load_lib`. The event when we call this method was changed. But not the role and the ownership of `_LIB` object.
> @trevor-m

Single instance of DLR is a limitation that we want to get rid of. We want to move to a stage where each model will come with its own local version of DLR, sitting in the same directory as the rest of the model artifacts. Consider having 2 models on the device, independent of each other, and an application which wants to run them concurrently (sequentially or in parallel, but both being loaded at the same time in application's memory). In this case, we will need to load each libdlr.so in its own namespace to keep them separate.

apivovarov:
This PR implements the following:
`libdlr.so` will be loaded one time when the first `DLRModelImpl` instance is created.
Once `libdlr.so` is loaded it will be used to run the first and all consequent models (even if they have their own `libdlr.so`).
Other `libdlr.so` (in other model folders) will not be used.

`_LIB` variable is initialized once for the entire Python process lifecycle.
Instead of initializing `_LIB` variable on `dlr import` we initialize it when the first `DLRModelImpl` instance is created. 
In this PR we only change the even which triggers `_LIB` initialization. There is no reason to move `_LIB` declaration to another hierarchical level. `_LIB` still has 1:1 relationship with DLR module. So, it should belong to the module. Initialization of _LIB happening when the first `DLRModelImpl` instance is created. Event which triggers `LIB` initialization is not the owner of `_LIB`. Owner of particular object is defined by the object relationships with other objects.
This is why `_LIB` belongs to module.

python/dlr/dlr_model.py
cnv1989: Resolved this offline.

PR #202, opened 2020-06-23
Move DLRModel declaration from dlr_common.h -> dlr_model.h.

cnv1989:
This PR contains housekeep changes to reduce the code footprint for refactor.

We will be deprecating support for Tensorflow and TFLite models going forward. This PR deletes all the C - Code related to Tensorflow and TFLite frameworks. The python implementation is independent of C++ and will be refactored along with rest of the python code.

The PR also moves DLRModel base class declaration from dlr_common.h to dlr_model.h to keep the declaration consistent with file names.

### DIFF
--- a/include/dlr_common.h
+++ b/include/dlr_common.h
@@ -5,8 +5,8 @@
 #include <dmlc/logging.h>
 #include <runtime_base.h>
 #include <sys/types.h>
-#include <nlohmann/json.hpp>
 
+#include <nlohmann/json.hpp>
 #include <string>
 #include <vector>
 
@@ -32,7 +32,7 @@ namespace dlr {
 
 /* The following file names are reserved by SageMaker and should not be used
  * as model JSON */
-constexpr const char* SAGEMAKER_AUXILIARY_JSON_FILES[] = {"model-shapes.json",
+constexpr const char *SAGEMAKER_AUXILIARY_JSON_FILES[] = {"model-shapes.json",
                                                           "hyperparams.json"};
 
 typedef struct {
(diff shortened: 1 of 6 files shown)


apivovarov:
In Java each class needs to be in its own java file with the same name. There is no such requirement in C++. `class DLRModel` can stay in `dlr_common.h`.

cnv1989:
> In Java each class needs to be in its own java file with the same name. There is no such requirement in C++. `class DLRModel` can stay in `dlr_common.h`. Do we have a case where we include `dlr_common.h` but not include `dlr_model.h`?

Thanks! Let me verify this.

cnv1989:
> In Java each class needs to be in its own java file with the same name. There is no such requirement in C++. `class DLRModel` can stay in `dlr_common.h`. Do we have a case where we include `dlr_common.h` but not include `dlr_model.h`?

I didn't understand the question? I this a suggestion? 

The reason for naming fix. Its hard to locate the base model. dlr_common.cc seems like a misleading name where one would expect to find common utilities. Its not a style specific to one language rather its a general best practice for any language. (https://medium.com/coding-skills/clean-code-101-meaningful-names-and-functions-bf450456d90c). 

dlr_common.cc is  doesn't indicate that it contains AbstractClass for all the models it better to keep it separate.

include/dlr_common.h
```diff
@@ -52,14 +51,12 @@ std::string GetParentFolder(const std::string& path);
...
-  return mainStr.size() >= toMatch.size() &&
-  mainStr.compare(0, toMatch.size(), toMatch) == 0;
+  return mainStr.size() >= toMatch.size() && mainStr.compare(0, toMatch.size(), toMatch) == 0;
 }
 
 inline bool EndsWith(const std::string& mainStr, const std::string& toMatch) {
-  if (mainStr.size() >= toMatch.size() &&
-      mainStr.compare(mainStr.size() - toMatch.size(), toMatch.size(),
-                      toMatch) == 0)
+  if (mainStr.size() >= toMatch.size() && \
```
trevor-m, line 58: `\` is not needed for extending line in C++

include/dlr_common.h
```diff
@@ -68,65 +65,11 @@ inline bool EndsWith(const std::string& mainStr, const std::string& toMatch) {
...
-  virtual std::vector<std::string> GetWeightNames() const = 0;
-  
-  virtual const char* GetBackend() const = 0;
-  virtual void SetNumThreads(int threads) = 0;
-  virtual bool HasMetadata() const {return false; }
-  virtual void UseCPUAffinity(bool use) = 0;
-  virtual void Run() = 0;
-  
-};
+CHECK_EQ(value, expected) << (msg) << ". Value read: " << (value) << ", Expected: " << (expected);
```
trevor-m, line 72: Is clang-format wanting to not indent this line?

cnv1989: thanks. Let me run it.

include/dlr_common.h
cnv1989: ran clang-format -style="google". Seem to have fixed it.

include/dlr_common.h
```diff
@@ -43,20 +43,20 @@ typedef struct {
   std::string metadata;
 } ModelPath;
 
-void ListDir(const std::string& dirname, std::vector<std::string>& paths);
+void ListDir(const std::string &dirname, std::vector<std::string> &paths);
```
trevor-m, line 46: FYI: having the `*` or `&` adjadcent to type or variable name are both acceptable: https://google.github.io/styleguide/cppguide.html#Pointer_and_Reference_Expressions

We don't have to change all of these. 
It seems like clang-format moved these automatically though?

cnv1989: Yeah!

PR #209, opened 2020-06-27
[Windows] Fix compilation errors

ymwangg:
These tweaks are needed to compile on Windows. cc @trevor-m @zhiics

### DIFF
--- a/include/dlr_tvm.h
+++ b/include/dlr_tvm.h
@@ -19,8 +19,7 @@ ModelPath GetTvmPaths(std::vector<std::string> tar_path);
 
 /*! \brief class TVMModel
  */
-
-class  DLR_DLL TVMModel : public DLRModel {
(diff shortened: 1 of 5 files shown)


include/dlr_tvm.h
```diff
@@ -13,7 +19,8 @@ ModelPath GetTvmPaths(std::vector<std::string> tar_path);
 
 /*! \brief class TVMModel
  */
-class TVMModel : public DLRModel {
+
+class  DLR_DLL TVMModel : public DLRModel {
```
trevor-m, line 23: Fix double space and extra newline

cnv1989: Perhaps run `clang-format -style="google"`

ymwangg: Good catch, fixed.

PR #212, opened 2020-07-02
Fix undefined behavior with GetDLRVersion.

trevor-m:
GetDLRVersion builds a new string to hold the version and then returns a pointer to it. The string goes out of scope when the function returns, so accessing the pointer is  undefined behavior.

Make version_str static so it persists after GetDLRVersion returns, so that the output pointer is still valid.

The current code can give errors like:
```
 ERROR error in DLRModel instantiation 'utf-8' codec can't decode byte 0x90 in position 0: invalid start byte
  File "/usr/local/lib/python3.6/dist-packages/dlr/api.py", line 82, in __init__
    self._impl = DLRModelImpl(model_path, dev_type, dev_id)
  File "/usr/local/lib/python3.6/dist-packages/dlr/dlr_model.py", line 118, in __init__
    self.version = self._get_version()
  File "/usr/local/lib/python3.6/dist-packages/dlr/dlr_model.py", line 97, in _get_version
    return version.value.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 0: invalid start byte
```

### DIFF
--- a/src/dlr.cc
+++ b/src/dlr.cc
@@ -329,9 +329,9 @@ extern "C" int GetDLRBackend(DLRModelHandle* handle, const char** name) {
 
 extern "C" int GetDLRVersion(const char** out) {
   API_BEGIN();
-  std::string version_str = std::to_string(DLR_MAJOR) + "." +
-                            std::to_string(DLR_MINOR) + "." +
-                            std::to_string(DLR_PATCH);
+  static const std::string version_str = std::to_string(DLR_MAJOR) + "." +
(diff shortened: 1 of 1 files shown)


lokeshgupta1975:
I suggest that since the version number for DLR is fixed at compile time, we need not evaluate it inside the function. Instead, we should define a global static std::string object in the main dlr.cc file, which will be known at compile time, and then the implementation of this function should be simple enough to just return a pointer to that string.

PR #223, opened 2020-08-03
Don't use system paths for libdlr when building wheel

trevor-m:
Currently when building the wheel or installing DLR, setup.py calls `find_lib_path` to locate the `libdlr.so` to include. However, `find_lib_path` prioritizes system directories over the build folder. This can cause the wrong dlr lib to be included in the wheel.

### DIFF
--- a/python/dlr/libpath.py
+++ b/python/dlr/libpath.py
@@ -10,20 +10,24 @@ class DLRLibraryNotFound(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


trevor-m:
@ wuchih-amazon @ymwangg  Could you review if these changes will work on windows?

trevor-m:
Windows only uses C apis and not the python wheel.

PR #236, opened 2020-09-17
Added GetInputShape & GetInputSizeDim Apis to dlr

cnv1989:


include/dlr.h
```diff
@@ -218,6 +218,32 @@ int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape,
  */
 DLR_DLL
 int GetDLRInput(DLRModelHandle* handle, const char* name, void* input);
+
+/*!
+ \brief Gets the shape of the index-th input.
+ \param handle The model handle returned from CreateDLRModel().
+ \param index The index-th input.
+ \param shape The pointer to save the shape of index-thinput. This should be a
```
trevor-m, line 226: Missing space between `index-thinput`

include/dlr.h
```diff
@@ -218,6 +218,32 @@ int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape,
...
+ pointer to an array of size "dim" from GetDLRInputSizeDim(). \return 0 for
```
trevor-m, line 227: \return should be on a new line

include/dlr_common.h
```diff
@@ -108,6 +109,9 @@ class DLR_DLL DLRModel {
   virtual int GetNumInputs() const { return num_inputs_; }
   virtual const char* GetInputName(int index) const = 0;
   virtual const char* GetInputType(int index) const = 0;
+  virtual const int GetInputDim(int index) const = 0;
+  virtual const int64_t GetInputSize(int index) const = 0;
+  virtual const std::vector<int64_t> &GetInputShape(int index) const;
```
trevor-m, line 114: `virtual const std::vector<int64_t>& GetInputShape(int index) const;`

src/dlr_treelite.cc
```diff
@@ -99,23 +113,30 @@ std::vector<std::string> TreeliteModel::GetWeightNames() const {
...
 const char* TreeliteModel::GetInputName(int index) const {
   CHECK_LT(index, num_inputs_) << "Input index is out of range.";
-  return "data";
+  return INPUT_NAME.c_str();
+}
+const int TreeliteModel::GetInputDim(int index) const { return kInputDim; }
+
+const int64_t TreeliteModel::GetInputSize(int index) const {
+  CHECK_LT(index, num_inputs_) << "Input index is out of range.";
+  std::vector<int64_t> shape = GetInputShape(index);
```
trevor-m, line 122: Make this a reference

src/dlr.cc
```diff
@@ -54,6 +54,26 @@ extern "C" int GetDLRInputType(DLRModelHandle* handle, int index,
   API_END();
 }
 
+extern "C" int GetDLRInputShape(DLRModelHandle* handle, int index,
+                                 int64_t* shape) {
+  API_BEGIN();
+  DLRModel* model = static_cast<DLRModel*>(*handle);
+  CHECK(model != nullptr) << "model is nullptr, create it first";
+  std::vector<int64_t> input_shape = model->GetInputShape(index);
```
trevor-m, line 62: Make `input_shape` a reference

src/dlr.cc
```diff
@@ -54,6 +54,26 @@ extern "C" int GetDLRInputType(DLRModelHandle* handle, int index,
...
+  std::memcpy(shape, input_shape.data(), sizeof(int64_t)*input_shape.size());
```
trevor-m, line 63: I think `std::copy(input_shape.begin(), input_shape.end(), shape);` is preferred here

src/dlr_common.cc
```diff
@@ -101,3 +101,8 @@ DLRBackend dlr::GetBackend(std::vector<std::string> dir_paths) {
   }
   return DLRBackend::kTREELITE;
 }
+
+const std::vector<int64_t>& DLRModel::GetInputShape(int index) const {
+  CHECK_LT(index, num_inputs_) << "Input index is out of range.";
+  return input_shapes_[index];
+}
```
trevor-m, line 108: Add newline

trevor-m (changes requested):
Thanks! Just some small comments

trevor-m:
Do we also need to make changes in the python code to use this API?

src/dlr.cc
cnv1989: GetInputShape returns a reference.

cnv1989:
> Do we also need to make changes in the python code to use this API?

It doesn't make sense to do it for python but I think we can do it to maintain consistency. I can do it as part of a follow up PR if thats fine.

src/dlr.cc
trevor-m: Yeah, so we need to do ` std::vector<int64_t>& input_shape = model->GetInputShape(index);` so that input_shape  is also a reference. Otherwise, the current code will create an unnecessary local copy of the shape.

src/dlr_relayvm.cc
```diff
@@ -240,16 +248,25 @@ void RelayVMModel::GetOutput(int index, void* output) {
 }
 
 void RelayVMModel::GetOutputShape(int index, int64_t* shape) const {
-  CHECK_LT(index, num_outputs_) << "Output index is out of range.";
-  std::memcpy(shape, outputs_[index]->shape, sizeof(int64_t) * outputs_[index]->ndim);
+  CHECK_LT(index, output_shapes_.size()) << "Output index is out of range.";
+  if (index < outputs_.size()) {
```
trevor-m, line 252: What are these two cases for? When would index be >= outputs.size()?

include/dlr_common.h
trevor-m: Can you fix the formatting here

trevor-m (changes requested):
Thanks for the fixes, I think we should add the python API as well, follow up PR is fine.

src/dlr_relayvm.cc
cnv1989: For dynamic models the output shapes are dynamic and depends on inference.  In the pre inference case the output shapes are obtained from metadata and after inference the output shapes are obtained from outputs.

src/dlr.cc
cnv1989: This change leads to `/home/ubuntu/TVM/neo-ai-dlr/src/dlr_treelite.cc:122:46: error: binding reference of type ‘std::vector<long int>&’ to ‘const std::vector<long int>’ discards qualifiers`

src/dlr_relayvm.cc
trevor-m: I see, so calling inference will initialize `outputs_`, otherwise it will use the metadata shapes from `output_shapes_`? To make it clearer, what do you think about doing something like:
```
if (outputs.empty()) {
  // Inference has not been called yet. Get shapes from metadata.
  CHECK_LT(index, output_shapes_.size()) << "Output index is out of range.";
  std::copy(output_shapes_[index].begin(), output_shapes_[index].end(), shape);
} else {
  CHECK_LT(index, outputs_.size()) << "Output index is out of range.";
  std::memcpy(shape, outputs_[index]->shape, sizeof(int64_t) * outputs_[index]->ndim);
}
```

trevor-m (approved):
Thanks!

apivovarov:
WITH_HEXAGON build is broken https://github.com/neo-ai/neo-ai-dlr/issues/251

PR #247, opened 2020-10-13
Delete tensorflow and tflite support from neo-ai-dlr.

cnv1989:

According to @ciphr we no longer need to support TF and TFLite in DLR. 

"We should not need to support TF-Lite as an execution runtime once we pull in the latest TVM that optimizes not only 176 out of 186 TF-Lite ops but also compiles int8 TFLite models. So TFLite models should be directly supported by TVM compilation. Second, we’re integrating TVM into TF so we should no longer need a separate TF runtime because models will execute in the TF/TVM inference-optimized-framework in the cloud. On edge devices, we would convert TF to TFLite and then compile to TVM. If a TF model cannot be converted to TFLite, customers will forgive us for not being able to compile that TF model to TVM (until TVM supports those TF operators directly).
 
Given this, I’m inclined to deprecate and discontinue support for TF and TFLite behind DLR. This was the original intent of this API – to provide a customer-opaque fall back to the original framework until TVM could compile the models from that framework." @ciphr


trevor-m:
Thanks @cnv1989 !

Can you also remove the following:
* TF/TFLite build instructions from `doc/install.rst`
* Running tf/tflite tests in Jenkinsfile
* Remove kTFLITE and kTENSORFLOW from DLRBackend enum `dlr_common.h` and remove both conditions for these in dlr::GetBackend `src/dlr_common.cc`.

cnv1989:
> ou also remove the following:
> 
> * TF/TFLite build instructions from `doc/install.rst`
> * Running tf/tflite tests in Jenkinsfile
> * Remove kTFLITE and kTENSORFLOW from DLRBackend enum `dlr_common.h` and remove both conditions for these in dlr::GetBackend `src/dlr_common.cc`.


Thanks 👍

apivovarov:
Do we still need include files? https://github.com/neo-ai/neo-ai-dlr/tree/master/include/dlr_tflite

trevor-m:
> Do we still need include files? https://github.com/neo-ai/neo-ai-dlr/tree/master/include/dlr_tflite

Good catch, we should delete that too @cnv1989

PR #249, opened 2020-10-14
Add more dlr.h functions to dlr_test

apivovarov:
Add more `dlr.h` functions to `dlr_test`
### Test output
```
[       OK ] DLR.TestRunDLRModel_GetDLROutput (268 ms)
[----------] 16 tests from DLR (3832 ms total)

[----------] Global test environment tear-down
[==========] 16 tests from 1 test case ran. (3832 ms total)
[  PASSED  ] 16 tests.
```

### DIFF
--- a/tests/cpp/dlr_test.cc
+++ b/tests/cpp/dlr_test.cc
@@ -24,6 +24,7 @@ TEST(DLR, TestGetDLRNumInputs) {
   int num_inputs;
   EXPECT_EQ(GetDLRNumInputs(&model, &num_inputs), 0);
   EXPECT_EQ(num_inputs, 1);
+  DeleteDLRModel(&model);
 }
 
 TEST(DLR, TestGetDLRNumWeights) {
@@ -31,6 +32,7 @@ TEST(DLR, TestGetDLRNumWeights) {
   int num_weights;
   EXPECT_EQ(GetDLRNumWeights(&model, &num_weights), 0);
   EXPECT_EQ(num_weights, 108);
+  DeleteDLRModel(&model);
 }
 
 TEST(DLR, TestGetDLRInputName) {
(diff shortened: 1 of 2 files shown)


cnv1989 (approved):

tests/cpp/dlr_test.cc
```diff
@@ -69,31 +74,48 @@ TEST(DLR, TestSetDLRInput) {
...
-
   int64_t output_size;
   int output_dim;
   int index = 0;
+  // output 0
   EXPECT_EQ(GetDLROutputSizeDim(&model, 0, &output_size, &output_dim), 0);
   EXPECT_EQ(output_dim, 1);
-  
+  EXPECT_EQ(output_size, 1);
+  int64_t shape0[output_dim];
```
trevor-m, line 107: Can you use `std::vector` or `new` to do this dynamic allocation? This will cause an error during windows build.

tests/cpp/dlr_test.cc
```diff
@@ -69,31 +74,48 @@ TEST(DLR, TestSetDLRInput) {
...
+  EXPECT_EQ(GetDLROutputShape(&model, 0, shape0), 0);
+  EXPECT_EQ(shape0[0], 1);
+  // output 1
   EXPECT_EQ(GetDLROutputSizeDim(&model, 1, &output_size, &output_dim), 0);
   EXPECT_EQ(output_dim, 2);
+  EXPECT_EQ(output_size, 1001);
+  int64_t shape1[output_dim];
```
trevor-m, line 114: `std::vector` or `new`

trevor-m (changes requested):
Thanks for adding these tests! 1 small change

trevor-m (approved):
Thanks @apivovarov !

tests/cpp/dlr_test.cc
apivovarov: Replaced it with `std::vector`. Btw, do we still support Mac build? I got `ld: unknown option: --exclude-libs` when building on  
- Mac OS10.14.6. 
- Apple clang version 11.0.0 (clang-1100.0.33.17).
- cmake version 3.12.1

tests/cpp/dlr_test.cc
apivovarov: fixed

PR #259, opened 2020-10-20
Add CreateDLRPipeline

apivovarov:
This PR adds `CreateDLRPipeline` to `dlr.h`

```
int CreateDLRPipeline(DLRModelHandle* handle,
                      int num_models,
                      const char** model_paths,
                      int dev_type,
                      int dev_id;
```

Resulting DLR model will execute its underlying models one after another

### DIFF
--- a/include/dlr_pipeline.h
+++ b/include/dlr_pipeline.h
@@ -1,0 +1,66 @@
+#ifndef DLR_PIPELINE_H_
+#define DLR_PIPELINE_H_
+
+#include <graph/graph_runtime.h>
+#include <tvm/runtime/memory.h>
+#include "dlr_common.h"
+
+#if defined(_MSC_VER) || defined(_WIN32)
+#define DLR_DLL __declspec(dllexport)
+#else
+#define DLR_DLL
+#endif  // defined(_MSC_VER) || defined(_WIN32)
+
+namespace dlr {
+
+/*! \brief class PipelineModel
+ */
+class DLR_DLL PipelineModel : public DLRModel {
+ private:
+  int count_;
+  const std::vector<DLRModelPtr> dlr_models_;
+  void CheckModelsCompatibility(const DLRModelPtr& m0, const DLRModelPtr& m1,
+                                const int m1_id, const bool is_runtime_check);
+  void SetupPipelineModel();
+
+ public:
+  /*! \brief Load model files from given folder path.
+   */
+  explicit PipelineModel(const std::vector<DLRModelPtr>& dlr_models, const DLContext& ctx)
+      : DLRModel(ctx, DLRBackend::kPIPELINE), dlr_models_(dlr_models) {
+    SetupPipelineModel();
+  }
+
+  virtual const int GetInputDim(int index) const override;
+  virtual const int64_t GetInputSize(int index) const override;
+  virtual const char* GetInputName(int index) const override;
+  virtual const char* GetInputType(int index) const override;
+  virtual void GetInput(const char* name, void* input) override;
+  virtual void SetInput(const char* name, const int64_t* shape, void* input,
+                        int dim) override;
+
+  virtual void GetOutput(int index, void* out) override;
+  virtual const void* GetOutputPtr(int index) const override;
+  virtual void GetOutputShape(int index, int64_t* shape) const override;
+  virtual void GetOutputSizeDim(int index, int64_t* size, int* dim) override;
+  virtual const char* GetOutputType(int index) const override;
+
+  virtual const char* GetWeightName(int index) const override;
+  virtual std::vector<std::string> GetWeightNames() const override;
+
+  virtual void Run() override;
+  virtual const char* GetBackend() const override;
+  virtual void SetNumThreads(int threads) override;
+  virtual void UseCPUAffinity(bool use) override;
+
+  /*
+    Following methods use metadata file to lookup input and output names.
+  */
+  virtual const char* GetOutputName(const int index) const override;
+  virtual int GetOutputIndex(const char* name) const override;
+  virtual void GetOutputByName(const char* name, void* out) override;
+};
+
+}  // namespace dlr
+
+#endif  // DLR_PIPELINE_H_
--- a/src/dlr.cc
+++ b/src/dlr.cc
@@ -1,9 +1,11 @@
 #include "dlr.h"
 
 #include "dlr_common.h"
+#include "dlr_pipeline.h"
+#include "dlr_relayvm.h"
 #include "dlr_treelite.h"
 #include "dlr_tvm.h"
-#include "dlr_relayvm.h"
+
 #ifdef DLR_HEXAGON
 #include "dlr_hexagon/dlr_hexagon.h"
 #endif  // DLR_HEXAGON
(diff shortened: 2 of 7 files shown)


include/dlr_pipeline.h
```diff
@@ -0,0 +1,64 @@
...
+class DLR_DLL PipelineModel : public DLRModel {
+ private:
+  int count_;
+  std::vector<DLRModelPtr> dlr_models_;
+  void SetupPipelineModel(std::vector<DLRModelPtr>& dlr_models);
+
+ public:
+  /*! \brief Load model files from given folder path.
+   */
+  explicit PipelineModel(std::vector<DLRModelPtr>& dlr_models, const DLContext& ctx)
```
trevor-m, line 27: Let's make `std::vector<DLRModelPtr>& dlr_models` -> `const std::vector<DLRModelPtr>& dlr_models`

Can we initialize `dlr_models_` here instead since we are just copying `dlr_models` to `dlr_models_`:
```
  : DLRModel(ctx, DLRBackend::kPIPELINE), dlr_models_(dlr_models) {
```

That way we don't need to pass to SetupPipelineModel

src/dlr_pipeline.cc
```diff
@@ -0,0 +1,162 @@
...
+#include <fstream>
+#include <iterator>
+#include <numeric>
+
+using namespace dlr;
+
+
+void PipelineModel::SetupPipelineModel(std::vector<DLRModelPtr>& dlr_models) {
+  CHECK_GT(dlr_models.size(), 0) << "List of models is empty";
+  dlr_models_ = dlr_models;
```
trevor-m, line 14: Let's operate only on the member copy `dlr_models_` in this function

src/dlr_pipeline.cc
```diff
@@ -0,0 +1,162 @@
...
+    input_names_.push_back(dlr_models[0]->GetInputName(i));
+    input_types_.push_back(dlr_models[0]->GetInputType(i));
+    input_shapes_.push_back(dlr_models[0]->GetInputShape(i));
+  }
+  // Check previous model outputs and current model inputs compatibility
+  for (int i = 1; i < count_; i++) {
+    DLRModelPtr pm = dlr_models[i - 1];
+    DLRModelPtr cm = dlr_models[i];
+    CHECK_EQ(pm->GetNumOutputs(), cm->GetNumInputs())
+      << "Number of outputs/inputs mismatch btw models "<< i - 1 << " and " << i
```
trevor-m, line 29: `btw` -> `between`

src/dlr_pipeline.cc
```diff
@@ -0,0 +1,162 @@
...
+  count_ = dlr_models.size();
+  num_inputs_ = dlr_models[0]->GetNumInputs();
+  num_weights_ = dlr_models[0]->GetNumWeights();
+  num_outputs_ = dlr_models.back()->GetNumOutputs();
+  for (int i = 0; i < num_inputs_; i++) {
+    input_names_.push_back(dlr_models[0]->GetInputName(i));
+    input_types_.push_back(dlr_models[0]->GetInputType(i));
+    input_shapes_.push_back(dlr_models[0]->GetInputShape(i));
+  }
+  // Check previous model outputs and current model inputs compatibility
```
trevor-m, line 24: For some relay VM models, the input might be unknown (`-1`) and the output dims could also unknown before running the model. Can you check for `-1` value in the shapes and bypass the size check in those cases?

trevor-m (changes requested):
Awesome, thanks @apivovarov ! Some small comments

include/dlr_pipeline.h
apivovarov: fixed

src/dlr_pipeline.cc
apivovarov: I added `PipelineModel::CheckModelsCompatibility()` method and use it in the constructor for initial models check and then in RUN method for output/input size/shape checks.

src/dlr_pipeline.cc
```diff
@@ -0,0 +1,182 @@
...
+  for (int j = 0; j < cm->GetNumInputs(); j++) {
+    int64_t out_size;
+    int out_dim;
+    pm->GetOutputSizeDim(j, &out_size, &out_dim);
+    if (!isRuntimeCheck) {
+      CHECK_EQ(out_dim, cm->GetInputDim(j))
+        << "Number of dimensions mismatch between output/input #" << j << ", models "
+        << cm_id - 1 << " and " << cm_id << std::endl;
+    }
+    const int64_t in_size = cm->GetInputSize(j);
```
trevor-m, line 29: We should change GetInputSize() to always return -1 for RelayVM if any one of the input shapes is -1. Currently, it is not guaranteed to be negative since it just multipleis all input shapes.

trevor-m (changes requested):
Thanks for fixing that! Just a very more small comments

src/dlr_pipeline.cc
```diff
@@ -0,0 +1,182 @@
...
+#include <stdlib.h>
+#include <cstring>
+#include <fstream>
+#include <iterator>
+#include <numeric>
+
+using namespace dlr;
+
+
+void PipelineModel::CheckModelsCompatibility(const DLRModelPtr& pm, const DLRModelPtr& cm,
```
cnv1989, line 12: Can we split this into two functions "ValidInput" and "ValidateOutputs"

include/dlr_pipeline.h
```diff
@@ -0,0 +1,66 @@
...
+#define DLR_DLL __declspec(dllexport)
+#else
+#define DLR_DLL
+#endif  // defined(_MSC_VER) || defined(_WIN32)
+
+namespace dlr {
+
+/*! \brief class PipelineModel
+ */
+class DLR_DLL PipelineModel : public DLRModel {
```
cnv1989, line 18: Nitpick: Instead of PipelineModel, should we call this "ModelPipeline" which in more representative of the underlying implementation and communicates whats actually happening.

src/dlr.cc
```diff
@@ -258,6 +281,30 @@ extern "C" int CreateDLRModel(DLRModelHandle* handle, const char* model_path,
...
+ */
+extern "C" int CreateDLRPipeline(DLRModelHandle* handle,
+                                 int num_models, const char** model_paths,
+                                 int dev_type, int dev_id) {
+  API_BEGIN();
+  DLContext ctx;
+  ctx.device_type = static_cast<DLDeviceType>(dev_type);
+  ctx.device_id = dev_id;
+  std::vector<DLRModelPtr> dlr_models;
+  for (int i = 0; i < num_models; i++) {
```
cnv1989, line 294: This logic should be implemented in CPP rather than here. Perhaps have a factory method or overloaded constructor that takes in paths vector.

src/dlr_pipeline.cc
apivovarov: opened new PR for that https://github.com/neo-ai/neo-ai-dlr/pull/259

include/dlr_pipeline.h
```diff
@@ -0,0 +1,66 @@
...
+ private:
+  int count_;
+  const std::vector<DLRModelPtr> dlr_models_;
+  void CheckModelsCompatibility(const DLRModelPtr& pm, const DLRModelPtr& cm,
```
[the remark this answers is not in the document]
apivovarov: - renamed to `is_runtime_check`
- CheckModelsCompatibility uses `m0` and `m1` as discussed.
- Setup and Run use `prev_model` and `curr_model`.

src/dlr_pipeline.cc
apivovarov: This function checks that m1.input matches m0.output

src/dlr.cc
apivovarov: CreateDLRPipeline body is only 15 lines. It is smaller that other Create methods... Also, it needs visibility to TVMModel, RelayVMModel, TreeliteModel and HexagonModel. The only file which has it is dlr.cc.

include/dlr_pipeline.h
apivovarov: Design doc says that in addition to `CreateDLRModel` we need new method `CreateDLRPipeline`.
Since all `DLRModel` classes ends with Model I named Pipeline class `PipelineModel`

trevor-m (approved):

PR #262, opened 2020-10-20
Input/Output Size should be -1 if shape has negative elements

apivovarov:
Input/Output Size should be -1 if shape has negative elements

### DIFF
--- a/src/dlr_common.cc
+++ b/src/dlr_common.cc
@@ -195,3 +195,7 @@ DLDeviceType dlr::GetDeviceTypeFromMetadata(const std::vector<std::string>& mode
     throw dmlc::Error("TargetDeviceType was not found in metadata!");
   } 
 }
+
+bool dlr::HasNegative(const int64_t* arr, const size_t size) {
+  return std::any_of(arr, arr + size, [](int64_t x) {return x < 0;});
+}
--- a/src/dlr_relayvm.cc
+++ b/src/dlr_relayvm.cc
@@ -153,6 +153,7 @@ const int RelayVMModel::GetInputDim(int index) const {
 
 const int64_t RelayVMModel::GetInputSize(int index) const {
   CHECK_LT(index, num_inputs_) << "Input index is out of range.";
+  if (dlr::HasNegative(input_shapes_[index].data(), input_shapes_[index].size())) return -1;
   return std::accumulate(input_shapes_[index].begin(), input_shapes_[index].end(), 1,
                          std::multiplies<int64_t>());
 }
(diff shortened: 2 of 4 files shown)


trevor-m (approved):
Nice, thanks @apivovarov !

cnv1989:
Let me check with smith to see if they are fine with this.

src/dlr_common.cc
```diff
@@ -195,3 +195,7 @@ DLDeviceType dlr::GetDeviceTypeFromMetadata(const std::vector<std::string>& mode
     throw dmlc::Error("TargetDeviceType was not found in metadata!");
   } 
 }
+
+bool dlr::HasNegative(const int64_t* arr, const size_t size) {
```
mycpuorg, line 199: nit:
```
dlr::shapeHasNegative()
```

cnv1989 (approved):

src/dlr_relayvm.cc
```diff
@@ -278,7 +279,11 @@ void RelayVMModel::GetOutputSizeDim(int index, int64_t* size, int* dim) {
   *size = 1;
   if (index < outputs_.size()) {
     auto arr = outputs_[index];
-    *size = std::accumulate(arr->shape, arr->shape + arr->ndim, 1, std::multiplies<int64_t>());
+    if (dlr::HasNegative(arr -> shape, arr -> ndim)) {
+      *size = -1;
```
mycpuorg, line 283: This could be an error code like
```
BAD_SIZE_ERR
```

src/dlr_tvm.cc
```diff
@@ -204,6 +205,10 @@ void TVMModel::GetOutputSizeDim(int index, int64_t* size, int* dim) {
   *size = 1;
   const DLTensor* tensor = outputs_[index];
   for (int i = 0; i < tensor->ndim; ++i) {
+    if (tensor->shape[i] < 0) {
```
mycpuorg, line 208: Overall, there is value in this change. Can we add unit tests for this change and add this to the documentation of the `GetOutputSizeDim()` and `GetInputDim()`

src/dlr_relayvm.cc
trevor-m: -1 is a valid size for RelayVM models, this PR is just so GetOutputSizeDim returns a consistent value for an unknown size

mycpuorg: I misunderstood your response:
All I was saying is:
```c++
constexpr int UNKNOWN_SIZE = -1;
```

PR #264, opened 2020-10-22
[DataTransform] Use kBadValue for invalid input float values

trevor-m:
For string input columns, unknown input strings are mapped to kMissingValue (-1).
For float input columns, invalid float inputs are now mapped to kBadValue (NaN) instead of -1, because -1 was a valid float.

### DIFF
--- a/src/dlr_data_transform.cc
+++ b/src/dlr_data_transform.cc
@@ -68,24 +68,29 @@ void DataTransform::MapToNDArray(const nlohmann::json& input_json,
   for (size_t r = 0; r < input_json.size(); ++r) {
     CHECK_EQ(input_json[r].size(), mapping.size()) << "Inconsistent number of columns";
     for (size_t c = 0; c < input_json[r].size(); ++c) {
-      try {
-        if (mapping[c].size()) {
-          // Look up in map. If not found, use kMissingValue.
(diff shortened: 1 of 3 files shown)


src/dlr_data_transform.cc
```diff
@@ -68,24 +68,29 @@ void DataTransform::MapToNDArray(const nlohmann::json& input_json,
   for (size_t r = 0; r < input_json.size(); ++r) {
     CHECK_EQ(input_json[r].size(), mapping.size()) << "Inconsistent number of columns";
     for (size_t c = 0; c < input_json[r].size(); ++c) {
-      try {
-        if (mapping[c].size()) {
```
apivovarov, line 72: Better to use boolean explicitly. I think `json` has `.empty()` method

src/dlr_data_transform.cc
```diff
@@ -68,24 +68,29 @@ void DataTransform::MapToNDArray(const nlohmann::json& input_json,
...
-          // Look up in map. If not found, use kMissingValue.
+      if (mapping[c].size()) {
```
apivovarov, line 71: Better to use boolean explicitly. I think json has .empty() method

apivovarov (approved):

PR #269, opened 2020-10-28
Add SetDLRCustomAllocator API.

trevor-m:
Supports custom allocation functions in DLR using `SetDLRCustomAllocator*()` functions.

Must be called before the model is loaded.
```
SetDLRCustomAllocatorMalloc(my_malloc);
SetDLRCustomAllocatorFree(my_free);
SetDLRCustomAllocatorMemalign(my_memalign);
CreateDLRModel(/* ... */);
```

### DIFF
--- a/include/dlr_allocator.h
+++ b/include/dlr_allocator.h
@@ -1,8 +1,7 @@
 #ifndef DLR_ALLOCATOR_H_
 #define DLR_ALLOCATOR_H_
 
-#include <dmlc/common.h>
-#include <dmlc/logging.h>
+#include <memory>
 
 #ifndef DLR_ALLOC_TYPEDEF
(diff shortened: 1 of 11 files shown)


include/dlr_allocator.h
```diff
@@ -0,0 +1,106 @@
+#ifndef DLR_ALLOCATOR_H_
+#define DLR_ALLOCATOR_H_
+
+#include <dmlc/common.h>
+#include <dmlc/logging.h>
```
apivovarov, line 5: Looks like `dlr_allocator.h` only needs `<memory>` header.
In addition to it `dlr_allocator.cc` also needs `<stdlib.h>`

apivovarov (approved):

PR #270, opened 2020-10-28
Use const void* for input

apivovarov:
- Mark input data pointer as const in C API `SetDLRInput` and C++ API `SetInput`. It is totally compatible with existing DLR API. No changes needed in existing users programs after DLR update.
- Change `LoadImageAndPreprocess` internal test utils method to return vector.
- Use vector to create input/output buffers in tests. (remove memalloc, new[], free and delete[])

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -223,7 +223,7 @@ int GetDLRWeightName(DLRModelHandle* handle, int index,
  */
 DLR_DLL
 int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape,
-                void* input, int dim);
+                const void* input, int dim);
 /*!
  \brief Gets the current value of the input according the node name.
  \param handle The model handle returned from CreateDLRModel().
--- a/tests/cpp/test_utils.hpp
+++ b/tests/cpp/test_utils.hpp
@@ -7,11 +7,11 @@
 #include <sstream>
 #include <cstring>
 
-float* LoadImageAndPreprocess(const std::string& img_path, size_t size,
-                              int batch_size) {
+std::vector<float> LoadImageAndPreprocess(const std::string& img_path, size_t size,
+                                          int batch_size) {
   std::string line;
   std::ifstream fp(img_path);
-  float* img = new float[size * batch_size];
+  std::vector<float> img(size * batch_size);
   size_t i = 0;
   if (fp.is_open()) {
     while (getline(fp, line) && i < size) {
@@ -26,7 +26,7 @@ float* LoadImageAndPreprocess(const std::string& img_path, size_t size,
   LOG(INFO) << "Image read - OK, float[" << i << "]";
 
   for (int j = 1; j < batch_size; j++) {
-    std::memcpy(img + j * size, img, size * sizeof(float));
+    std::copy_n(img.cbegin(), size, img.begin() + j * size);
   }
   return img;
 }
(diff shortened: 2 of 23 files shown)


trevor-m:
Will this change break ABI?
We shouldn’t change existing DRL C APIs unless necessary

apivovarov:
This change will not break ABI because users can still use `void*` variables for `SetDLRInput `. No errors / no warnings even with `-Wall -Wextra`.
```
int i = 5;
int *a = &i;
int const *b = a;  /* Totally fine. No warnings */
```

apivovarov:
in `dlr_dlsym_test` I tested that:
- dlsym can load `SetDLRInput` function using old non-const function signature.
- new `SetDLRInput` (with const input) can accept const and non-const input pointers

tests/cpp/test_utils.hpp
```diff
@@ -26,7 +26,7 @@ float* LoadImageAndPreprocess(const std::string& img_path, size_t size,
   LOG(INFO) << "Image read - OK, float[" << i << "]";
 
   for (int j = 1; j < batch_size; j++) {
-    std::memcpy(img + j * size, img, size * sizeof(float));
+    std::copy_n(img.cbegin(), size, img.begin() + j * size);
```
mycpuorg, line 29: Nice.

include/dlr.h
```diff
@@ -223,7 +223,7 @@ int GetDLRWeightName(DLRModelHandle* handle, int index,
  */
 DLR_DLL
 int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape,
-                void* input, int dim);
+                const void* input, int dim);
```
mycpuorg, line 226: We get away only because `dlsym` returns a `void*`

What this PR does is less risky, since losing the `const`-ness is what's usually flagged by compilers.
However, it is important to ensure the backward compat of a stored type info is not broken, say, I have the following in my existing DLR client application code in `C++`.

```C++
// dlrHandle = dlopen();
using foo_ptr_t = int (*)(void*, const char*, const int64_t*, void*, int);
auto my_fptr = dlsym(dlrHandle, "SetDLRInput");

// my_set_in_ptr is void *
((foo_ptr_t)my_fptr)(...);  // -> does this throw an error or warning?
```

It's good to mention this case was tested explicitly (casting old pointer type on the new one)

trevor-m (approved):

include/dlr.h
apivovarov: Manoj, `dlr_dlsym_test.cc` tests that old SetDLRInput (with non-const input) still works.
```
int (*SetDLRInput)(DLRModelHandle* handle, const char* name, const int64_t* shape,
                   void* input, int dim);
// New signature of SetDLRInput - input data pointer marked as const
int (*SetDLRInputC)(DLRModelHandle* handle, const char* name, const int64_t* shape,
                    const void* input, int dim);
                    
*(void**)(&SetDLRInput) = dlsym(dlr, "SetDLRInput");
*(void**)(&SetDLRInputC) = dlsym(dlr, "SetDLRInput");

float* img_data = img.data();
const float* cimg_data = img.data();

// Test that old SetDLRInput (with non-const input) still works
EXPECT_EQ(SetDLRInput(&model, input_name, shape, img_data, 4), 0);
// Test that new SetDLRInput accepts const and non-const inputs
EXPECT_EQ(SetDLRInputC(&model, input_name, shape, img_data, 4), 0);
EXPECT_EQ(SetDLRInputC(&model, input_name, shape, cimg_data, 4), 0);
```

mycpuorg: Thanks for quick turn around.

LGTM.

mycpuorg (approved):
LGTM

mycpuorg (approved):

PR #272, opened 2020-10-31
Fix to allow building on macOS

### LINKED ISSUE #261: Building on Mac OS X
Hello,
I am following the steps mentioned at: https://neo-ai-dlr.readthedocs.io/en/latest/install.html#building-on-mac-os-x to build `dlr` on my mac.  I am trying to install the version 1.3.0 of dlr. However, building on the latest branch is also failing with the same error. 

`make -j4` command fails with the following output 
```
[ 25%] Linking CXX shared library libgtest.dylib
ld: unknown option: --exclude-libs
collect2: error: ld returned 1 exit status
make[2]: *** [googletest/googletest-build/googlemock/gtest/libgtest.dylib] Error 1
make[1]: *** [googletest/googletest-build/googlemock/gtest/CMakeFiles/gtest.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
......
[ 75%] Linking CXX static library libtvm_runtime_static.a
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libtvm_runtime_static.a(rpc_pipe_impl.cc.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libtvm_runtime_static.a(rpc_pipe_impl.cc.o) has no symbols
[ 75%] Built target tvm_runtime_static
```

The behavior is same with any `gcc` version over 8. The `cmake` version that I am using is `3.18.4`

Upon explicitly setting `export CMAKE_SYSTEM_NAME=darwin`, the build continues to 81% and errors out with the same  error. 

```
ld: unknown option: --exclude-libs
collect2: error: ld returned 1 exit status
make[2]: *** [googletest/googletest-build/googlemock/gtest/libgtest.dylib] Error 1
make[1]: *** [googletest/googletest-build/googlemock/gtest/CMakeFiles/gtest.dir/all] Error 2 
```

TylerADavis:
This PR implements the fix proposed by saranyailla in #261 

Currently includes `-t` as an additional linker arg. If this is removed, then we should be able to merge the `WIN32` and `APPLE` checks.

I've verified that this builds without issue on macOS.

Fixes #261.


apivovarov:
- for Windows we do not append `--exclude-libs,ALL`
- for Mac we append `-t` (according to `man ld` - 
```
-t   Logs each file (object, archive, or dylib) the linker loads.  Useful for debugging problems with search paths where the wrong library is loaded
```
- for other cases we append `--exclude-libs,ALL`

I have couple questions
Why we need `-t` for Mac?
What if we make Mac case similar to Windows case - just do not append `--exclude-libs,ALL`.

Looks like `--exclude-libs,ALL` is needed to avoid conflicts in case when we load multiple `libdlr.so` at the runtime.
```
from dlr import DLRModel
model1 = DLRModel(model_path1, 'cpu')   # model folder has its own libdlr.so
model2 = DLRModel(model_path2, 'cpu').  # model folder has its own libdlr.so
```
Error: `Global PackedFunc runtime.module.loadfile_so is already registered`

TylerADavis:
Thanks for the feedback @apivovarov . I've updated `CMakeLists.txt` with a comment along the lines of what you mentioned, and I've also combined the Windows and Mac logic together.

The `-t` was likely originally in place for debugging, so it was fine to cut.

I've verified the new changes build without issue on Mac.

apivovarov (approved):

PR #276, opened 2020-11-06
DLR APIs for get/set DLTensor and bug fixes for creating models from files

samskalicky:
## Description
Adds new APIs for DLTensor support and bug fixes for creating models from files:
- Fixes CreateDLRModel for specifying paths directly to model files (lib.so, graph.json, etc...) instead of a directory
- Adds APIs for supporting DLTensor in setting inputs/outputs 

While all backends benefit from the path fixes, the new DLTensor support is only for TVM GraphRuntime and VMRuntime backends for these new APIs.

### SetDLRInputTensor
Allows setting input by providing a DLTensor directly instead of passing separate data, shape, dim.
- Added support for both GraphRuntime and VMRuntime, other backend types are not supported (checks added with appropriate error messages)
- Added corresponding `SetInputTensor` APIs on `TVMModel` and `RelayVMModel` classes
- For `RelayVMModel` eventually calls `CopyFrom` for the input NDArray or runs transform
- For `TVMModel` eventually calls `SetInput` on the graphRuntime object

### GetDLROutputTensor
Allows getting an output by copying a from TVM outputs to a given pre-allocated DLTensor.
- Added support for both GraphRuntime and VMRuntime, other backend types are not supported (checks added with appropriate error messages)
- Added corresponding `GetOutputTensor` APIs on `TVMModel` and `RelayVMModel` classes
- For `RelayVMModel` eventually calls `GetOutput` if a transform is set, or `CopyTo` for the output NDArray
- For `TVMModel` eventually calls `get_output` on graphRuntime object

### GetDLROutputManagedTensorPtr
Returns a pointer to a DLManagedTensor for a particular output.
- Added support for both GraphRuntime and VMRuntime, other backend types are not supported (checks added with appropriate error messages)
- Added corresponding `GetOutputManagedTensorPtr` APIs on `TVMModel` and `RelayVMModel` classes
- For `RelayVMModel` eventually calls `ToDLPack` for the output NDArray, if a transform is set an error is thrown since theres no way to get a `DLManagedTensor` from the transform
- For `TVMModel` eventually calls `get_output` to get the output NDArray and then `ToDLPack`

## Other changes
- refactored `FixWindowsDriveLetter` functionality from multiple places (MakePathVec, GetDLRDeviceType) into a single function
  - Also fixed bug in that was only fixing the path for the first entry, now loops over all paths
- [fixed the `ListDir` function](https://github.com/neo-ai/neo-ai-dlr/pull/276#discussion_r520742182) to handle paths to specific files rather than directories. 
- refactored each of the backends to take a list of paths to the files they need (so, params, json, meta, ro, etc...). 
  - This deduplicated the code that was searching the path multiple times: 
    - getBackend
    - GetTvmPaths
    - GetHexagonModelFile
    - InitModelPath
    - GetTreelitePaths
  - Added a new `FindFiles` function to do this once and save the files found in the search in a vector and just reuse those in `CreateDLRModelPtr` in a single place.
- removed `ver_json` field from ModelPath struct for version.json file that is [no longer used anywhere](https://github.com/neo-ai/neo-ai-dlr/blob/3571813712412b38ef5b086fcc6cdf088f7b613a/src/dlr_treelite.cc#L88)
- refactored `GetBackend` in `DLRModel` class to use enum instead of a string to avoid string compares when checking model type
- Refactored each of the backends setup of `ModelPath` struct into `InitModelPath` in dlr_common this deduplicated the code that was checking for the same metadata/lib.so files.
- Added check to detect multiple duplicate model files and error out accordingly (ie. multiple lib.so's)

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -48,25 +48,6 @@ typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
 #endif
 
 /*!
- \brief TVM model file paths for GraphRuntime or VMRuntime
- */
-typedef struct TVMPaths_t {
-  TVMPaths_t()
-      : model_lib(NULL),
-        params(NULL),
-        model_json(NULL),
-        ver_json(NULL),
-        metadata(NULL),
-        relay_executable(NULL) {}
-  const char* model_lib;         // Required for TVM GraphRuntime & VMRuntime
-  const char* params;            // Required for TVM GraphRUntime
-  const char* model_json;        // Required for TVM GraphRuntime
-  const char* ver_json;          // Optional
-  const char* metadata;          // Required for TVM VMRuntime
-  const char* relay_executable;  // Required for TVM VMRuntime
-} TVMPaths;
-
-/*!
  \brief Creates a DLR model by searching in given directory path.
  \param handle The pointer to save the model handle.
  \param model_path Path to the folder containing the model files,
@@ -484,19 +465,6 @@ int SetDLRCustomAllocatorMemalign(DLRMemalignFunctionPtr custom_memalign_fn);
 DLR_DLL
 int CreateTVMModelFromMemory(DLRModelHandle* handle, const char* graph, const char* lib_path,
                              const char* params, size_t params_len, int dev_type, int dev_id);
-
-/*!
- * \brief Creates a DLR model for TVM GraphRuntime and VMRuntime models from specified paths
- *        to individual files rather than a directory.
- * \param handle The pointer to save the model handle.
- * \param paths Paths to the model files stored in different locations
- * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
- * \param dev_id Device ID.
- * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateTVMModelFromPaths(DLRModelHandle* handle, const TVMPaths* paths, int dev_type,
-                            int dev_id);
 
 /*!
  * \brief Sets the input according the node name from existing DLTensor. Can only be
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -72,43 +72,6 @@ DLR_DLL
 int CreateDLRModelFromHexagon(DLRModelHandle* handle, const char* model_path, int debug_level);
 #endif  // DLR_HEXAGON
 
-/*
- * \brief Creates a DLR model for TVM GraphRuntime models. Accepts graph.json, params data,
- *        and metadata (optional) from memory rather than loading from files on disk.
- *        The lib.so must still be on disk since dlopen will be used to load it.
- * \param handle The pointer to save the model handle.
- * \param lib_path Path to the TVM lib.so file on disk
- * \param graph String of loaded graph.json file
- * \param params Binary data loaded from TVM params file
- * \param params_len length of the params data
- * \param metadata file contents, or empty string (cannot be nullptr, 0)
- * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
- * \param dev_id Device ID.
- * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateDLRModelFromGraphRuntime(DLRModelHandle* handle, const char* lib_path, const char* graph,
-                                   const char* params, size_t params_len, const char* metadata,
-                                   int dev_type, int dev_id);
-
-/*
- * \brief Creates a DLR model for TVM VMRuntime models. Accepts relay executable and metadata
- *        from memory rather than loading from files on disk. lib.so must still be on disk
- *        since dlopen will be used to load it.
- * \param handle The pointer to save the model handle.
- * \param lib_path Path to the TVM lib.so file on disk
- * \param relay_exec Binary data loaded from relay executable file code.ro
- * \param relay_len length of the relay executable file data
- * \param metadata file contents, or empty string (cannot be nullptr, 0)
- * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
- * \param dev_id Device ID.
- * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateDLRModelFromVMRuntime(DLRModelHandle* handle, const char* lib_path,
-                                const char* relay_exec, size_t relay_len, const char* metadata,
-                                int dev_type, int dev_id);
-
 /*!
  \brief Creates a DLR pipeline model.
  \param handle The pointer to save the model handle.
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -59,66 +59,6 @@ typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
 DLR_DLL
 int CreateDLRModel(DLRModelHandle* handle, const char* model_path, int dev_type, int dev_id);
 
-#ifdef DLR_TFLITE
-/*!
- \brief Creates a DLR model from TFLite
- \param handle The pointer to save the model handle.
- \param model_path Path to tflite file or to the top-level directory containing
- tflite file \param threads Number of threads to use. \param use_nnapi Use
- NNAPI, 0 - false, 1 - true. \return 0 for success, -1 for error. Call
- DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateDLRModelFromTFLite(DLRModelHandle* handle, const char* model_path, int threads,
-                             int use_nnapi);
-#endif  // DLR_TFLITE
-
-#ifdef DLR_TENSORFLOW
-
-/*!
- \brief Tensorflow GPU Options structure for Tensorflow Config structure.
- */
-typedef struct DLR_TFGPUOptions {
-  int allow_growth;
-  double per_process_gpu_memory_fraction;
-} DLR_TFGPUOptions;
-
-/*!
- \brief Tensorflow Config structure for CreateDLRModelFromTensorflow.
- */
-typedef struct DLR_TFConfig {
-  int intra_op_parallelism_threads;
-  int inter_op_parallelism_threads;
-  DLR_TFGPUOptions gpu_options;
-} DLR_TFConfig;
-
-/*!
- \brief Tensor Description structure for CreateDLRModelFromTensorflow.
- */
-typedef struct DLR_TFTensorDesc {
-  const char* name;
-  const int64_t* dims;
-  const int num_dims;
-} DLR_TFTensorDesc;
-
-/*!
- \brief Creates a DLR model from Tensorflow frozen model .pb file
- \param handle The pointer to save the model handle.
- \param model_path Path to .pb file or to the top-level directory containing .pb
- file \param inputs array of input names \param input_size size of input array
- \param outputs array of output names
- \param output_size size of output array
- \param batch_size set batch size for the case when the model has dynamic batch
- size \param threads Number of threads to use (ignored if zero of less) \return
- 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateDLRModelFromTensorflow(DLRModelHandle* handle, const char* model_path,
-                                 const DLR_TFTensorDesc* inputs, int input_size,
-                                 const char* outputs[], int output_size,
-                                 const DLR_TFConfig tf_config);
-#endif  // DLR_TENSORFLOW
-
 #ifdef DLR_HEXAGON
 /*!
  \brief Creates a DLR model from Hexagon
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -431,7 +431,7 @@ int SetDLRInputTensor(DLRModelHandle* handle, const char* name, void* dltensor);
  * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int GetDLROutputManagedTensorPtr(DLRModelHandle* handle, int index, void** dlmanagedtensor);
+int GetDLROutputManagedTensorPtr(DLRModelHandle* handle, int index, const void** dlmanagedtensor);
 
 /*!
  * \brief Gets the index-th output from the model and copies it into the given DLTensor.
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -399,13 +399,16 @@ int SetDLRCustomAllocatorMemalign(DLRMemalignFunctionPtr custom_memalign_fn);
  * \param lib_path Path to the TVM lib.so file on disk
  * \param params Binary data loaded from TVM params file
  * \param params_len length of the params data
+ * \param metadata file contents
+ * \param version file contents
  * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
  * \param dev_id Device ID.
  * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int CreateTVMModelFromMemory(DLRModelHandle* handle, const char* graph, const char* lib_path,
-                             const char* params, size_t params_len, int dev_type, int dev_id);
+int CreateDLRModelFromGraphRuntime(DLRModelHandle* handle, const char* lib_path, const char* graph,
+                                   const char* params, size_t params_len, const char* metadata,
+                                   int dev_type, int dev_id);
 
 /*!
  * \brief Sets the input according the node name from existing DLTensor. Can only be
@@ -416,18 +419,19 @@ int CreateTVMModelFromMemory(DLRModelHandle* handle, const char* graph, const ch
  * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int SetTVMInputTensor(DLRModelHandle* handle, const char* name, void* dltensor);
+int SetDLRInputTensor(DLRModelHandle* handle, const char* name, void* dltensor);
 
 /*!
  * \brief Gets the index-th output from the model and sets the pointer to it.
  *        Can only be used with TVM models (GraphRuntime and VMRuntime)
  * \param handle The model handle returned from CreateDLRModel().
  * \param index The index-th output.
- * \param out The pointer to a DLManagedTensor pointer, will be set.
+ * \param out The pointer to an unallocated DLManagedTensor pointer, will be
+ *            set by this function to point to an internal DLManagedTensor.
  * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int GetTVMOutputManagedTensor(DLRModelHandle* handle, int index, void** dlmanagedtensor);
+int GetDLROutputManagedTensorPtr(DLRModelHandle* handle, int index, void** dlmanagedtensor);
 
 /*!
  * \brief Gets the index-th output from the model and copies it into the given DLTensor.
@@ -438,7 +442,7 @@ int GetTVMOutputManagedTensor(DLRModelHandle* handle, int index, void** dlmanage
  * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int CopyTVMOutputTensor(DLRModelHandle* handle, int index, void* dltensor);
+int GetDLROutputTensor(DLRModelHandle* handle, int index, void* dltensor);
 
 /*! \} */
 
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -2,6 +2,7 @@
 #define DLR_H_
 
 #include <stdint.h>
+#include <stddef.h>
 
 /* special symbols for DLL library on Windows */
 #ifdef __cplusplus
@@ -37,21 +38,22 @@ extern "C" {  // Open extern "C" block
 typedef void* DLRModelHandle;
 
 /*!
- \brief Paths for model files
+ \brief TVM model file paths for GraphRuntime or VMRuntime
  */
-typedef struct DLRPaths_t {
-  DLRPaths_t()
-      : model_lib(0), params(0), model_json(0), ver_json(0), metadata(0), relay_executable(0) {}
-  const char* model_lib;
-  const char* params;
-  const char* model_json;
-  const char* ver_json;
-  const char* metadata;
-  const char* relay_executable;
-} DLRPaths;
+typedef struct TVMPaths_t {
+  TVMPaths_t()
+      : model_lib(NULL), params(NULL), model_json(NULL), ver_json(NULL),
+    metadata(NULL), relay_executable(NULL) {}
+  const char* model_lib;         // Required for TVM GraphRuntime & VMRuntime
+  const char* params;            // Required for TVM GraphRUntime
+  const char* model_json;        // Required for TVM GraphRuntime
+  const char* ver_json;          // Optional
+  const char* metadata;          // Required for TVM VMRuntime
+  const char* relay_executable;  // Required for TVM VMRuntime
+} TVMPaths;
 
 /*!
- \brief Creates a DLR model by search in directory path given.
+ \brief Creates a DLR model by searching in given directory path.
  \param handle The pointer to save the model handle.
  \param model_path Path to the folder containing the model files,
                    or colon-separated list of folders (or files) if model files
@@ -61,33 +63,6 @@ typedef struct DLRPaths_t {
  */
 DLR_DLL
 int CreateDLRModel(DLRModelHandle* handle, const char* model_path, int dev_type, int dev_id);
-
-/*!
- * \brief Creates a TVM DLR model.
- * \param handle The pointer to save the model handle.
- * \param graph String of loaded graph.json file
- * \param lib_path Path to the TVM lib.so file
- * \param params String of binary data loaded from tvm.params file
- * \param params_len length of the params string
- * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
- * \param dev_id Device ID.
- * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateTVMModel(DLRModelHandle* handle, const char* graph, const char* lib_path,
-                   const char* params, unsigned params_len, int dev_type, int dev_id);
-
-/*!
- * \brief Creates a DLR model from specified paths
- * \param handle The pointer to save the model handle.
- * \param paths Paths to the model files stored in different locations
- * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
- * \param dev_id Device ID.
- * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int CreateDLRModelFromPaths(DLRModelHandle* handle, const DLRPaths* paths, int dev_type,
-                            int dev_id);
 
 #ifdef DLR_TFLITE
 /*!
@@ -261,15 +236,6 @@ DLR_DLL
 int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape, const void* input,
                 int dim);
 
-/*!
- * \brief Sets the input according the node name.
- * \param handle The model handle returned from CreateDLRModel().
- * \param name The input node name.
- * \param tensor The input DLTensor.
- * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
- */
-DLR_DLL
-int SetTVMInputTensor(DLRModelHandle* handle, const char* name, void* dltensor);
 /*!
  \brief Gets the current value of the input according the node name.
  \param handle The model handle returned from CreateDLRModel().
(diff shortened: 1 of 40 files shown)


include/dlr.h
```diff
@@ -50,6 +61,34 @@ DLR_DLL
...
+/*!
+ \brief Creates a DLR model from specified paths
+ \param handle The pointer to save the model handle.
+ \param paths Paths to the model files
+ stored in different locations \param dev_type Device type. Valid values are in
+ the DLDeviceType enum in dlpack.h. \param dev_id Device ID. \return 0 for
+ success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int CreateDLRModelFromPaths(DLRModelHandle* handle, const DLRPaths* paths, int dev_type,
```
kkoppolu1, line 89: `DLRPaths` can be a const ref

include/dlr_relayvm.h
```diff
@@ -45,6 +45,7 @@ class DLR_DLL RelayVMModel : public DLRModel {
   std::vector<std::vector<int64_t>> output_shapes_;
   DataTransform data_transform_;
   void InitModelPath(std::vector<std::string> paths);
+  void InitModelPath(ModelPath paths);
```
kkoppolu1, line 48: This is probably not an appropriate constructor because not all attributes of `ModelPath` are applicable to RelayVM.
The factory method creating the various backends should be the one inspecting the struct and invoking the constructors of appropriate backends with the correct arguments.

src/dlr.cc
```diff
@@ -281,6 +300,71 @@ extern "C" int CreateDLRModel(DLRModelHandle* handle, const char* model_path,
...
+  path.model_lib = paths->model_lib;
+  path.params = paths->params;
+  path.model_json = paths->model_json;
+  path.ver_json = paths->ver_json;
+  path.metadata = paths->metadata;
+  path.relay_executable = paths->relay_executable;
+  
+  DLRModel* model;
+  try {
+    if (paths->model_lib != NULL and paths->params != NULL and paths->model_json != NULL) {
```
kkoppolu1, line 322: Is `and` a C/C++ keyword? Is it identical to `&&` ?

samskalicky: doh....was clearly writing python just before this................

include/dlr.h
```diff
@@ -50,6 +61,34 @@ DLR_DLL
...
+ \param dev_id Device ID. 
+ \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int CreateTVMModel(DLRModelHandle* handle, const char* graph,
+                   const char* lib_path, const char* params, unsigned params_len,
+                   int dev_type, int dev_id);
+  
+/*!
+ \brief Creates a DLR model from specified paths
```
trevor-m, line 81: We can format the documentation like this (add * to beginning of each line) to avoid the formatting getting messed up.
```
/*!
 * \brief Creates a DLR model from specified paths
 * \param handle The pointer to save the model handle.
 * \param paths Paths to the model files stored in different locations
 * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
 * \param dev_id Device ID.
 * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
 */

You only need to do this for the functions you added

**Review by trevor-m — CHANGES_REQUESTED** [review]
Thanks Sam! Could you also apply clang-format to the files you've edited.

File: include/dlr_relayvm.h (hunk 2 above)
Comment (samskalicky) [reply include/dlr_relayvm.h:48]: Thats ok, it shares `model_lib` and `metadata` fields in `ModelPath` with GraphRuntime. Plus we're just doing the same thing that the other `InitModelPath` does:
https://github.com/neo-ai/neo-ai-dlr/blob/5bb6d49c80a2733feb1187fb3a6077b9567f21b7/src/dlr_relayvm.cc#L14-L28

**Comment by trevor-m** [comment]
Also, could you create unit tests in `tests/cpp` for the new APIs?

File: src/dlr_tvm.cc [hunk 5]
```diff
@@ -164,6 +173,31 @@ void TVMModel::SetInput(const char* name, const int64_t* shape,
   UpdateInputShapes();
 }
 
+void TVMModel::SetInput(const char* name, DLTensor* tensor) {
+  std::string str(name);
+  int index = tvm_graph_runtime_->GetInputIndex(str);
+  if(index > -1) {
+    tvm::runtime::NDArray arr = tvm_graph_runtime_->GetInput(index);
+    DLTensor input_tensor = *(arr.operator->());
+    CHECK(tensor->ctx.device_type == input_tensor.ctx.device_type)
```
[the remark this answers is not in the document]
samskalicky: Should we change it to a warning? I would like to know if im doing something inefficient thats causing cross-device copies.

include/dlr.h
```diff
@@ -37,8 +37,19 @@ extern "C" {  // Open extern "C" block
  */
 typedef void* DLRModelHandle;
 
-/*!
- \brief Creates a DLR model.
+typedef struct DLRPaths_t {
+DLRPaths_t() : model_lib(0), params(0), model_json(0),
+    ver_json(0), metadata(0), relay_executable(0) {}
+  char* model_lib;
```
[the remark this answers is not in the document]
samskalicky: done

include/dlr_tvm.h
```diff
@@ -27,6 +27,8 @@ class DLR_DLL TVMModel : public DLRModel {
   std::vector<std::string> output_types_;
   std::vector<std::string> weight_names_;
   void SetupTVMModule(std::vector<std::string> model_path);
+  void SetupTVMModule(ModelPath paths);
+  void SetupTVMModule(std::string json_str, std::string param_str, ModelPath paths);
```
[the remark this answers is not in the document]
samskalicky: done

src/dlr.cc
```diff
@@ -281,6 +300,71 @@ extern "C" int CreateDLRModel(DLRModelHandle* handle, const char* model_path,
...
+                              const char* lib_path,
+                              const char* params,
+                              unsigned params_len,
+                              int dev_type, int dev_id) {
+  API_BEGIN();
+  DLContext ctx;
+  ctx.device_type = static_cast<DLDeviceType>(dev_type);
+  ctx.device_id = dev_id;
+  /* Logic to handle Windows drive letter */
+  auto path_fix = [=](auto path) {
```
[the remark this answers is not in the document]
samskalicky: done

include/dlr.h
samskalicky: done

samskalicky:
ran clang-format

include/dlr_tvm.h
```diff
@@ -27,8 +27,8 @@ class DLR_DLL TVMModel : public DLRModel {
   std::vector<std::string> output_types_;
   std::vector<std::string> weight_names_;
   void SetupTVMModule(std::vector<std::string> model_path);
-  void SetupTVMModule(ModelPath paths);
-  void SetupTVMModule(std::string json_str, std::string param_str, ModelPath paths);
+  void SetupTVMModule(const ModelPath& paths);
+  void SetupTVMModule(const std::string& json_str, const std::string& param_str, const ModelPath& paths);
```
trevor-m: Could you make `SetupTVMModule` take a `dmlc::Stream` instead of string for param_str?

That way it can handle this change in the custom allocator PR:
https://github.com/neo-ai/neo-ai-dlr/pull/269/files#diff-0c87eebc446be9132c7f9bae311fa7b6b339bac8d650e23e7569b9561f8c7fdaR79-R80

When you call SetupTVMModule, for now you can use
`SetupTVMModule(json_str, dmcl::MemoryStringStream(&param_str), paths);`

include/dlr_tvm.h
```diff
@@ -27,6 +27,9 @@ class DLR_DLL TVMModel : public DLRModel {
   std::vector<std::string> output_types_;
   std::vector<std::string> weight_names_;
   void SetupTVMModule(std::vector<std::string> model_path);
+  void SetupTVMModule(const ModelPath& paths);
+  void SetupTVMModule(const std::string& json_str, const std::string& param_str,
```
trevor-m, line 31: Could you make `SetupTVMModule` take a `dmlc::Stream` instead of string for param_str?

That way it can handle this change in the custom allocator PR:
https://github.com/neo-ai/neo-ai-dlr/pull/269/files#diff-0c87eebc446be9132c7f9bae311fa7b6b339bac8d650e23e7569b9561f8c7fdaR79-R80

When you call SetupTVMModule, for now you can use
`SetupTVMModule(json_str, dmcl::MemoryStringStream(&param_str), paths);`

trevor-m (changes requested):

include/dlr_common.h
```diff
@@ -55,6 +54,9 @@ typedef struct {
 
 void ListDir(const std::string& dirname, std::vector<std::string>& paths);
 
+/* Logic to handle Windows drive letter */
+std::string pathFix(std::string path);
```
trevor-m, line 58: Can you rename to something more descriptive like `FixWindowsDriveLetter`

trevor-m (changes requested):

src/dlr_tvm.cc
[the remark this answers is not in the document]
trevor-m: I see, I don't think it would be good to have a warning everytime it is called. Maybe we can add a new api to get the device type for an input (or just get the DLTensor)? That way the application can do the check.

include/dlr_common.h
samskalicky: done

include/dlr_tvm.h
samskalicky: done

include/dlr_tvm.h
samskalicky: done

src/dlr_tvm.cc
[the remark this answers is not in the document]
samskalicky: we do this already here:
https://github.com/neo-ai/neo-ai-dlr/blob/139ab3089f9eb9639522c701a3158755e2212c88/src/dlr_tvm.cc#L156-L161

src/dlr.cc
```diff
@@ -70,35 +66,41 @@ extern "C" int GetDLRInputSizeDim(DLRModelHandle* handle, int index,
...
   DLRModel* model = static_cast<DLRModel*>(*handle);
   CHECK(model != nullptr) << "model is nullptr, create it first";
   model->SetInput(name, shape, input, dim);
   API_END();
 }
 
-extern "C" int GetDLRInput(DLRModelHandle* handle, const char* name,
-                           void* input) {
+extern "C" int SetTVMInputTensor(DLRModelHandle* handle, const char* name, void* dltensor) {
+  API_BEGIN();
```
trevor-m, line 87: Can you add a check here and to other new TVM model specific functions to make sure DLRBackend == kTVM?

include/dlr.h
```diff
@@ -47,8 +60,34 @@ typedef void* DLRModelHandle;
...
+ * \param graph String of loaded graph.json file
+ * \param lib_path Path to the TVM lib.so file
+ * \param params String of binary data loaded from tvm.params file
+ * \param params_len length of the params string
+ * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
+ * \param dev_id Device ID.
+ * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int CreateTVMModel(DLRModelHandle* handle, const char* graph, const char* lib_path,
```
trevor-m, line 77: Can you use `size_t` type instead of `unsigned params_len`?

Add `#include <stddef.h>`

include/dlr.h
```diff
@@ -47,8 +60,34 @@ typedef void* DLRModelHandle;
  success, -1 for error. Call DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int CreateDLRModel(DLRModelHandle* handle, const char* model_path, int dev_type,
-                   int dev_id);
+int CreateDLRModel(DLRModelHandle* handle, const char* model_path, int dev_type, int dev_id);
+
+/*!
+ * \brief Creates a TVM DLR model.
```
trevor-m, line 66: Can you expand a bit more in the docstring on the difference between this and a regular DLR model? Maybe clarify this is for advanced users.

Can we also group all the TVM specific APIs together, perhaps at the end of the file? Most regular users probably won't uses these and it might be confusing.

include/dlr.h
```diff
@@ -222,8 +258,18 @@ int GetDLRWeightName(DLRModelHandle* handle, int index,
  message.
  */
 DLR_DLL
-int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape,
-                const void* input, int dim);
+int SetDLRInput(DLRModelHandle* handle, const char* name, const int64_t* shape, const void* input,
+                int dim);
+
+/*!
+ * \brief Sets the input according the node name.
```
trevor-m, line 265: Mention how this can only be used for TVM graph runtime models (such as those created by CreateTVMModel)

include/dlr.h
```diff
@@ -324,8 +368,17 @@ int GetDLROutputSizeDim(DLRModelHandle* handle, int index, int64_t* size,
...
+
+/*!
+ * \brief Gets the index-th output from the model.
+ * \param handle The model handle returned from CreateDLRModel().
+ * \param index The index-th output.
+ * \param out The pointer to save the output DLTensor
+ * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int GetTVMOutputTensor(DLRModelHandle* handle, int index, void* dltensor);
```
trevor-m, line 381: Mention how this can only be used for TVM graph runtime models (such as those created by CreateTVMModel)

trevor-m (changes requested):
Thanks for the fixes! Some remaining concerns.

We also need
* Add tests for new APIs. Test new C apis in dlr_test.cc. You can also unit tests for C++ TVMModel changes in dlr_tvm_test.cc.

include/dlr.h
trevor-m: Mention that these only work when `GetDLRBackend` returns tvm

include/dlr.h
samskalicky: done

include/dlr.h
samskalicky: done

include/dlr.h
samskalicky: done

include/dlr.h
samskalicky: done

src/dlr.cc
samskalicky: done, used strcmp since GetBackend returns a const char*. Currently theres no API get a `DLRBackend` enum value. If we decide to change this I can update the check to use that

include/dlr.h
```diff
@@ -60,8 +78,8 @@ int CreateDLRModel(DLRModelHandle* handle, const char* model_path, int dev_type,
  DLRGetLastError() to get the error message.
  */
 DLR_DLL
-int CreateDLRModelFromTFLite(DLRModelHandle* handle, const char* model_path,
-                             int threads, int use_nnapi);
+int CreateDLRModelFromTFLite(DLRModelHandle* handle, const char* model_path, int threads,
```
apivovarov, line 81: Do we still need FromTFLite?

samskalicky: im not changing it here, just ran clang-format per @trevor-m 's instructions

samskalicky:
ill add some tests for VMruntime flow tomorrow. then we should be done

include/dlr.h
```diff
@@ -416,6 +428,69 @@ int SetDLRNumThreads(DLRModelHandle* handle, int threads);
...
+ * \brief Creates a DLR model for TVM GraphRuntime and VMRuntime models from specified paths
+ *        to individual files rather than a directory.
+ * \param handle The pointer to save the model handle.
+ * \param paths Paths to the model files stored in different locations
+ * \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
+ * \param dev_id Device ID.
+ * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int CreateTVMModelFromPaths(DLRModelHandle* handle, const TVMPaths* paths, int dev_type,
```
trevor-m, line 458: CreateDLRModel also accepts colon separated paths to files:
`model_path="/foo/model.so;/bar/model.json;/buzz/model.params"`.
Can we use that increase of creating a new API? 

@apivovarov

samskalicky: the issue is the pathvec only contains the first path in the list. When we get the colon separated string:
https://github.com/neo-ai/neo-ai-dlr/blob/3a6d2276d00ed89f5c620a18e56947e64d4c5611/src/dlr.cc#L209-L211
We only take the first one and add it to the pathvec:
https://github.com/neo-ai/neo-ai-dlr/blob/3a6d2276d00ed89f5c620a18e56947e64d4c5611/src/dlr.cc#L204-L206

samskalicky: if you pass multiple paths like: `./resnet_v1_5_50/compiled.so:./resnet_v1_5_50/compiled_model.json:./resnet_v1_5_50/compiled.params`
you get this error:
```
[00:24:29] /home/ubuntu/dlr/tests/cpp/dlr_test.cc:240: TVMError: LocalFileSystem.ListDirectory ./resnet_v1_5_50/compiled.so error: Not a directory
```
Since the ListDir call just assumes each entry in the path is a directory:
https://github.com/neo-ai/neo-ai-dlr/blob/3a6d2276d00ed89f5c620a18e56947e64d4c5611/src/dlr_common.cc#L50-L54
and then it fails when it realizes the entry is a file and not a directory
https://github.com/dmlc/dmlc-core/blob/bfad207b448480783a1f428ae3d93d87032d8349/src/io/local_filesys.cc#L101-L105

samskalicky: Theres no API to check if a path is a directory in dmlc or tvm or dlr

src/dlr.cc
```diff
@@ -181,38 +181,29 @@ extern "C" int GetDLROutputByName(DLRModelHandle* handle, const char* name, void
...
 #ifdef DLR_HEXAGON
   } else if (backend == DLRBackend::kHEXAGON) {
-    const std::string model_path_string(model_path);
-    return std::make_shared<HexagonModel>(model_path_string, ctx, 1 /*debug_level*/);
+    return std::make_shared<HexagonModel>(files, ctx, 1 /*debug_level*/);
 #endif  // DLR_HEXAGON
   } else {
-    throw dmlc::Error("Unsupported backend!");
+    LOG(ERROR) << "Unable to determine backend from path: '" << model_path << "'.";
+    return nullptr;  // unreachable
```
trevor-m, line 206: LOG(ERROR) isnt fatal (that's LOG(FATAL)) so this is actually reachable. Can you instead throw dmlc::Error? dmlc::Error will be caught higher up and cause the C API function to return -1.

src/dlr_common.cc
```diff
@@ -68,26 +84,30 @@ void dlr::LoadJsonFromFile(const std::string& path, nlohmann::json& jsonObject)
...
-  for (auto dir : dir_paths) {
-    dlr::ListDir(dir, paths);
-  }
-  for (auto filename : paths) {
+  return files;
+}
+
+DLRBackend dlr::GetBackend(const std::vector<std::string>& files) {
+  // Scan files to guess the backend.
+  bool hasSO = false;
```
trevor-m, line 97: Rename to `has_so` or `has_lib` to follow naming convention

src/dlr_relayvm.cc
```diff
@@ -237,6 +232,27 @@ void RelayVMModel::SetInput(const char* name, const int64_t* shape, const void*
   inputs_[index] = input_arr;
 }
 
+void RelayVMModel::SetInput(const char* name, DLTensor* tensor) {
```
trevor-m, line 235: Why is this showing up in diff? Did it get duplicated?

include/dlr_common.h
```diff
@@ -56,6 +56,11 @@ typedef struct {
 
 void ListDir(const std::string& dirname, std::vector<std::string>& paths);
 
+std::vector<std::string> FindFiles(std::vector<std::string> paths);
```
trevor-m, line 59: use const vector& for args

include/dlr_common.h
```diff
@@ -56,6 +56,11 @@ typedef struct {
...
+
+/* Logic to handle Windows drive letter */
+std::string FixWindowsDriveLetter(std::string path);
```
trevor-m, line 62: use const string& for args

src/dlr.cc
```diff
@@ -243,23 +234,23 @@ extern "C" int CreateDLRModel(DLRModelHandle* handle, const char* model_path, in
   ctx.device_id = dev_id;
 
   std::vector<std::string> path_vec = MakePathVec(model_path);
+  std::vector<std::string> files = FindFiles(path_vec);
 
-  DLRBackend backend = dlr::GetBackend(path_vec);
+  DLRBackend backend = dlr::GetBackend(files);
   DLRModel* model;
   try {
     if (backend == DLRBackend::kTVM) {
```
trevor-m, line 242: Could you replace this block with CreateDLRModelPtr()? Sorry we should've done that earlier

trevor-m (changes requested):
Thanks @samskalicky
Everything looks good, just some minor nit-picky comments now

src/dlr_relayvm.cc
samskalicky: this is a new API. Its different from the existing one:
https://github.com/neo-ai/neo-ai-dlr/blob/3571813712412b38ef5b086fcc6cdf088f7b613a/src/dlr_relayvm.cc#L213

src/dlr.cc
samskalicky: I dont think its that simple. CreateDLRModelPtr allocates on the stack, but CreateDLRModel allocates with new

include/dlr_common.h
samskalicky: done

include/dlr_common.h
samskalicky: done

src/dlr_common.cc
samskalicky: done

src/dlr.cc
samskalicky: done

src/dlr_relayvm.cc
trevor-m: I see, thanks!

trevor-m (approved):
Thank you Sam!

@apivovarov Could you take a look again?

include/dlr.h
```diff
@@ -48,7 +48,7 @@ typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
 #endif
 
 /*!
- \brief Creates a DLR model.
+ \brief Creates a DLR model by searching in given directory path.
```
apivovarov, line 51: It is not only a directory might be given, but colon separated list of files.

samskalicky: updated

include/dlr.h
```diff
@@ -388,6 +390,56 @@ int SetDLRCustomAllocatorFree(DLRFreeFunctionPtr custom_free_fn);
...
 
+/*
+ * \brief Creates a DLR model for TVM GraphRuntime models. Accepts graph.json and params data
+ *        from memory rather than loading from files on disk. lib.so must still be on disk
+ *        since dlopen will be used to load it.
+ * \param handle The pointer to save the model handle.
+ * \param graph String of loaded graph.json file
+ * \param lib_path Path to the TVM lib.so file on disk
+ * \param params String of binary data loaded from TVM params file
+ * \param params_len length of the params string
```
apivovarov, line 401: Because we are asking users to provide `params_len` I guess the array of chars might not have terminating null character.
In that case we should not call it "String".  It is just a byte array.

samskalicky: thats right, it may not be null terminated, and in fact might have "null characters" in the middle since its just binary data.

src/dlr.cc
```diff
@@ -381,3 +365,94 @@ extern "C" int SetDLRCustomAllocatorMemalign(DLRMemalignFunctionPtr custom_memal
...
+extern "C" int SetDLRInputTensor(DLRModelHandle* handle, const char* name, void* dltensor) {
+  API_BEGIN();
+  DLRModel* dlr_model = static_cast<DLRModel*>(*handle);
+  CHECK(strcmp(dlr_model->GetBackend(), "tvm") == 0 ||
+        strcmp(dlr_model->GetBackend(), "relayvm") == 0)
+      << "model is not a TVMModel or RelayVMModel. Found '" << dlr_model->GetBackend()
+      << "' but expected 'tvm' or 'relayvm'";
+
+  DLTensor* tensor = static_cast<DLTensor*>(dltensor);
+  if (strcmp(dlr_model->GetBackend(), "tvm") == 0) {
```
kkoppolu1, line 405: nit:  use strncmp to handle non-null terminated strings safely

src/dlr_common.cc
```diff
@@ -50,15 +50,31 @@ std::string dlr::GetBasename(const std::string& path) {
 void dlr::ListDir(const std::string& dirname, std::vector<std::string>& paths) {
   dmlc::io::URI uri(dirname.c_str());
```
kkoppolu1, line 51: change the variable name since its no longer just a directory name

src/dlr.cc
samskalicky: they are null terminated: 
https://github.com/neo-ai/neo-ai-dlr/blob/master/src/dlr_tvm.cc#L237
And there are lots of other strcmp uses elsewhere. Do we need to change this now or can we do it globally for DLR in another PR?

src/dlr_common.cc
samskalicky: changed dirname to path

src/dlr.cc
kkoppolu1: Follow up is fine

include/dlr.h
```diff
@@ -388,6 +390,77 @@ int SetDLRCustomAllocatorFree(DLRFreeFunctionPtr custom_free_fn);
...
+ */
+DLR_DLL
+int SetDLRInputTensor(DLRModelHandle* handle, const char* name, void* dltensor);
+
+/*!
+ * \brief Gets the index-th output from the model and sets the pointer to it.
+ *        Can only be used with TVM models (GraphRuntime and VMRuntime)
+ * \param handle The model handle returned from CreateDLRModel().
+ * \param index The index-th output.
+ * \param out The pointer to an unallocated DLManagedTensor pointer, will be
```
apivovarov, line 446: Looks like it it `dlmanagedtensor` but not 'out'. BTW, can we use short name `tensor`.  it is only one tensor param in the function and the name of the method already says that type of tensor it is `GetDLROutputManagedTensorPtr`

include/dlr.h
```diff
@@ -388,6 +390,77 @@ int SetDLRCustomAllocatorFree(DLRFreeFunctionPtr custom_free_fn);
...
+/*!
+ * \brief Gets the index-th output from the model and copies it into the given DLTensor.
+ *        Can only be used with TVM models (GraphRuntime and VMRuntime)
+ * \param handle The model handle returned from CreateDLRModel().
+ * \param index The index-th output.
+ * \param out The pointer to an existing/allocated DLTensor to copy the output into.
+ * \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int GetDLROutputTensor(DLRModelHandle* handle, int index, void* dltensor);
```
apivovarov, line 462: Same comments about `out` in docs. 
Can we put Getters before Setters.  And put `GetDLROutputTensor` before `GetDLROutputManagedTensorPtr`

samskalicky: done

include/dlr.h
samskalicky: done

kkoppolu1:
.

kkoppolu1 (approved):

src/dlr.cc
samskalicky: @kkoppolu1 i just pushed a change that uses enum instead of strings for this logic. so we're addressing your comment in an other way.

kkoppolu1: nice!

include/dlr_common.h
```diff
@@ -76,11 +81,12 @@ inline bool EndsWith(const std::string& mainStr, const std::string& toMatch) {
     return false;
 }
 
-enum class DLRBackend { kTVM, kTREELITE, kHEXAGON, kRELAYVM, kPIPELINE };
+enum class DLRBackend { kTVM, kTREELITE, kHEXAGON, kRELAYVM, kPIPELINE, kUNKNOWN };
+extern const char* backend2str[6];
```
trevor-m, line 85: `kBackendToStr`

samskalicky: done

src/dlr_common.cc
```diff
@@ -105,20 +105,51 @@ std::vector<std::string> dlr::FindFiles(const std::vector<std::string>& paths) {
...
+        std::string msg = "Found multiple *_hexagon_model.so files: ";
+        msg += hexagon + " " + filename;
+        throw dmlc::Error(msg);
+      }
+      hexagon = filename;
       return DLRBackend::kHEXAGON;
     } else if (EndsWith(filename, ".so")) {
-      has_so = true;  // dont return immediately since so could be part of many diff backend types
+      if (lib.length() > 0) {
+        std::string msg = "Found multiple *.so files: ";
```
trevor-m, line 135: Files already excludes libdlr.so right?

samskalicky: oops, will update to match https://github.com/neo-ai/neo-ai-dlr/blob/master/src/dlr_tvm.cc#L26

samskalicky: refactored, done

trevor-m (approved):

PR #284, opened 2020-11-16
[DataTranform] Make output transform more robust

trevor-m:
Current code would crash if `GetOutputShape` or `GetOutputSizeDim` was called before inference was ran for a model with an output DataTransform. Now DLR will give shape of -1.


trevor-m:
https://github.com/neo-ai/neo-ai-dlr/pull/321

PR #291, opened 2020-11-18
Add CreateDLRModelFromModelElem

apivovarov:
This PR adds `CreateDLRModelFromModelElem` function.
Model Element can be file path or pointer to data in memory

Example:
```
  std::vector<DLRModelElem> model_elems = {
    {DLRModelElemType::TVM_GRAPH, nullptr, graph_str.c_str(), 0},
    {DLRModelElemType::TVM_PARAMS, nullptr, params_str.data(), params_str.size()},
    {DLRModelElemType::TVM_LIB, so_file.c_str(), nullptr, 0},
    {DLRModelElemType::NEO_METADATA, meta_file.c_str(), nullptr, 0}
  };

CreateDLRModelFromModelElem(&model, model_elems.data(), model_elems.size(), device_type, 0)
```

### DIFF
--- a/include/dlr.h
+++ b/include/dlr.h
@@ -47,6 +47,17 @@ typedef void (*DLRFreeFunctionPtr)(void*);
 typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
 #endif
 
+#ifndef DLR_MODEL_ELEM
+#define DLR_MODEL_ELEM
+enum DLRModelElemType { HEXAGON_LIB, NEO_METADATA, TVM_GRAPH, TVM_LIB, TVM_PARAMS, RELAY_EXEC };
+typedef struct ModelElem {
+  const DLRModelElemType type;
+  const char* path;
+  const void* data;
+  const size_t data_size;
+} DLRModelElem;
+#endif
+
 /*!
  \brief Creates a DLR model.
  \param handle The pointer to save the model handle.
@@ -58,6 +69,18 @@ typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
  */
 DLR_DLL
 int CreateDLRModel(DLRModelHandle* handle, const char* model_path, int dev_type, int dev_id);
+
+/*!
+ \brief Creates a DLR model from model elements.
+ \param handle The pointer to save the model handle.
+ \param model_elems DLR Model elements. Element can be file path or data pointer in memory.
+ \param dev_type Device type. Valid values are in the DLDeviceType enum in dlpack.h.
+ \param dev_id Device ID.
+ \return 0 for success, -1 for error. Call DLRGetLastError() to get the error message.
+ */
+DLR_DLL
+int CreateDLRModelFromModelElem(DLRModelHandle* handle, const DLRModelElem* model_elems,
+                                size_t model_elems_size, int dev_type, int dev_id);
 
 #ifdef DLR_HEXAGON
 /*!
--- a/include/dlr_common.h
+++ b/include/dlr_common.h
@@ -75,6 +75,8 @@ std::string GetParentFolder(const std::string& path);
 
 void LoadJsonFromFile(const std::string& path, nlohmann::json& jsonObject);
 
+void LoadJsonFromString(const std::string& jsonData, nlohmann::json& jsonObject);
+
 inline bool StartsWith(const std::string& mainStr, const std::string& toMatch) {
   return mainStr.size() >= toMatch.size() && mainStr.compare(0, toMatch.size(), toMatch) == 0;
 }
--- a/include/dlr_common.h
+++ b/include/dlr_common.h
@@ -77,6 +77,9 @@ void LoadJsonFromFile(const std::string& path, nlohmann::json& jsonObject);
 
 void LoadJsonFromString(const std::string& jsonData, nlohmann::json& jsonObject);
 
+std::string LoadFileToString(const std::string& path,
+                             std::ios_base::openmode mode = std::ios_base::in);
+
 inline bool StartsWith(const std::string& mainStr, const std::string& toMatch) {
   return mainStr.size() >= toMatch.size() && mainStr.compare(0, toMatch.size(), toMatch) == 0;
 }
--- a/include/dlr_common.h
+++ b/include/dlr_common.h
@@ -39,6 +39,17 @@
 #define DLR_DLL
 #endif  // defined(_MSC_VER) || defined(_WIN32)
 
+#ifndef DLR_MODEL_ELEM
+#define DLR_MODEL_ELEM
+enum DLRModelElemType { HEXAGON_LIB, NEO_METADATA, TVM_GRAPH, TVM_LIB, TVM_PARAMS, RELAY_EXEC };
+typedef struct ModelElem {
+  const DLRModelElemType type;
+  const char* path;
+  const void* data;
+  const size_t data_size;
+} DLRModelElem;
+#endif
+
 namespace dlr {
 
 /* The following file names are reserved by SageMaker and should not be used
@@ -76,11 +87,13 @@ inline bool EndsWith(const std::string& mainStr, const std::string& toMatch) {
     return false;
 }
 
-enum class DLRBackend { kTVM, kTREELITE, kHEXAGON, kRELAYVM, kPIPELINE };
+enum class DLRBackend { kTVM, kTREELITE, kHEXAGON, kRELAYVM, kPIPELINE, kUNKNOWN };
 
 /*! \brief Get the backend based on the contents of the model folder.
  */
-DLRBackend GetBackend(std::vector<std::string> dirname);
+DLRBackend GetBackend(const std::vector<std::string>& dirname);
+
+DLRBackend GetBackend(const std::vector<DLRModelElem>& model_elems);
 
 std::string GetMetadataFile(const std::string& dirname);
 
--- a/src/dlr.cc
+++ b/src/dlr.cc
@@ -271,6 +271,35 @@ extern "C" int CreateDLRModel(DLRModelHandle* handle, const char* model_path, in
   API_END();
 }
 
+extern "C" int CreateDLRModelFromModelElem(DLRModelHandle* handle, const DLRModelElem* model_elems,
+                                           size_t model_elems_size, int dev_type, int dev_id) {
+  API_BEGIN();
+  DLContext ctx;
+  ctx.device_type = static_cast<DLDeviceType>(dev_type);
+  ctx.device_id = dev_id;
+
+  std::vector<DLRModelElem> model_elem_vec(model_elems, model_elems + model_elems_size);
+
+  DLRBackend backend = dlr::GetBackend(model_elem_vec);
+  DLRModel* model;
+  try {
+    if (backend == DLRBackend::kTVM) {
+      model = new TVMModel(model_elem_vec, ctx);
+    } else if (backend == DLRBackend::kRELAYVM) {
+      model = new RelayVMModel(model_elem_vec, ctx);
+    } else {
+      LOG(FATAL) << "Unsupported backend!";
+      return -1;  // unreachable
+    }
+  } catch (dmlc::Error& e) {
+    LOG(ERROR) << e.what();
+    return -1;
+  }
+
+  *handle = model;
+  API_END();
+}
+
 /*! \brief Translate c args from ctypes to std types for DLRModel ctor.
  */
 extern "C" int CreateDLRPipeline(DLRModelHandle* handle, int num_models, const char** model_paths,
--- a/src/dlr_common.cc
+++ b/src/dlr_common.cc
@@ -77,6 +77,13 @@ void dlr::LoadJsonFromString(const std::string& jsonData, nlohmann::json& jsonOb
   }
 }
 
+std::string dlr::LoadFileToString(const std::string& path, std::ios_base::openmode mode) {
+  std::ifstream fstream(path, mode);
+  std::stringstream blob;
+  blob << fstream.rdbuf();
+  return blob.str();
+}
+
 DLRBackend dlr::GetBackend(const std::vector<std::string>& dir_paths) {
   // Support the case where user provides full path to hexagon file.
   if (EndsWith(dir_paths[0], "_hexagon_model.so")) {
--- a/src/dlr_common.cc
+++ b/src/dlr_common.cc
@@ -68,7 +68,7 @@ void dlr::LoadJsonFromFile(const std::string& path, nlohmann::json& jsonObject) 
   }
 }
 
-DLRBackend dlr::GetBackend(std::vector<std::string> dir_paths) {
+DLRBackend dlr::GetBackend(const std::vector<std::string>& dir_paths) {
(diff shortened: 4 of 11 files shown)


include/dlr.h
```diff
@@ -47,6 +47,17 @@ typedef void (*DLRFreeFunctionPtr)(void*);
 typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
 #endif
 
+#ifndef DLR_MODEL_ELEM
+#define DLR_MODEL_ELEM
+enum DLRModelElemType { HEXAGON_LIB, NEO_METADATA, TVM_GRAPH, TVM_LIB, TVM_PARAMS, RELAY_EXEC };
+typedef struct ModelElem {
+   const DLRModelElemType type;
```
samskalicky, line 54: should we make this a string (or char*) so that when we have new file types we dont need to update the enum?

include/dlr_common.h
```diff
@@ -39,6 +39,17 @@
...
+#ifndef DLR_MODEL_ELEM
+#define DLR_MODEL_ELEM
+enum DLRModelElemType { HEXAGON_LIB, NEO_METADATA, TVM_GRAPH, TVM_LIB, TVM_PARAMS, RELAY_EXEC };
+typedef struct ModelElem {
+   const DLRModelElemType type;
+   const char* path;
+   const void* data;
+   const size_t data_size;
+} DLRModelElem;
+#endif
```
samskalicky, line 51: can we deduplicate this in dlr.h?

src/dlr.cc
```diff
@@ -271,6 +271,35 @@ extern "C" int CreateDLRModel(DLRModelHandle* handle, const char* model_path, in
...
+  std::vector<DLRModelElem> model_elem_vec(model_elems, model_elems + model_elems_size);
+
+  DLRBackend backend = dlr::GetBackend(model_elem_vec);
+  DLRModel* model;
+  try {
+    if (backend == DLRBackend::kTVM) {
+      model = new TVMModel(model_elem_vec, ctx);
+    } else if (backend == DLRBackend::kRELAYVM) {
+      model = new RelayVMModel(model_elem_vec, ctx);
+    } else {
```
samskalicky, line 290: are we planning to support treelite/hexagon later?

src/dlr_tvm.cc
```diff
@@ -57,27 +70,68 @@ void TVMModel::SetupTVMModule(std::vector<std::string> model_path) {
...
   tvm::runtime::Module module;
-  if (!IsFileEmpty(paths.model_lib)) {
-    module = tvm::runtime::Module::LoadFromFile(paths.model_lib);
+  if (!IsFileEmpty(model_lib_path)) {
+    module = tvm::runtime::Module::LoadFromFile(model_lib_path);
   }
-  if (!paths.metadata.empty() && !IsFileEmpty(paths.metadata)) {
-    LoadJsonFromFile(paths.metadata, this->metadata_);
+  if (!metadata_path.empty() && !IsFileEmpty(metadata_path)) {
+    LoadJsonFromFile(metadata_path, this->metadata_);
```
samskalicky, line 128: Should we merge the other PR first and then use one of these SetupTVMModule functions to deduplicate the code for setting up the graphRuntime object?
https://github.com/neo-ai/neo-ai-dlr/pull/276/files#diff-0c87eebc446be9132c7f9bae311fa7b6b339bac8d650e23e7569b9561f8c7fdaR74

include/dlr.h
apivovarov: The common practice is to use string constants or enums for enumerated keys. if we add the support new type then we will add unit tests anyway. Adding new enum element should not create any issues.

include/dlr_common.h
apivovarov: In DLR we have few external C types and many internal C++ types. C++ backend does not see external C types. In rare cases internal and external declarations are the same. e.g. typedef `DLRMallocFunctionPtr`.  In that case we declare them in both dlr.h (external) and dlr_common.h (internal)

src/dlr.cc
apivovarov: It is totally possible.

src/dlr_tvm.cc
apivovarov: I'd prefer to merge this PR first. After that we will refactor PR 276 into 2-3 smaller PRs.

src/dlr_relayvm.cc
```diff
@@ -48,10 +59,44 @@ void RelayVMModel::SetupVMModule() {
...
+    if (el.type == DLRModelElemType::RELAY_EXEC) {
+      if (el.path != nullptr) {
+        std::ifstream relay_ob(el.path, std::ios::binary);
+        std::string code_data_((std::istreambuf_iterator<char>(relay_ob)),
+                               std::istreambuf_iterator<char>());
+        code_data = code_data_;
+      } else if (el.data != nullptr && el.data_size > 0) {
+        const char* code_data_cstr = static_cast<const char*>(el.data);
+        std::string code_data_(code_data_cstr, el.data_size);
+        code_data = code_data_;
```
trevor-m, line 76: Can do this all in one line with `code_data.assign(static_cast<const char*>(el.data), el.data_size)`

src/dlr_relayvm.cc
```diff
@@ -48,10 +59,44 @@ void RelayVMModel::SetupVMModule() {
...
+  std::string code_data;
+  std::string model_lib_path;
+  std::string metadata_path;
+  for (DLRModelElem el: model_elems) {
+    if (el.type == DLRModelElemType::RELAY_EXEC) {
+      if (el.path != nullptr) {
+        std::ifstream relay_ob(el.path, std::ios::binary);
+        std::string code_data_((std::istreambuf_iterator<char>(relay_ob)),
+                               std::istreambuf_iterator<char>());
+        code_data = code_data_;
```
trevor-m, line 72: We can use `code_data.assign((std::istreambuf_iterator<char>(relay_ob)),
                               std::istreambuf_iterator<char>());`

assign is basically like calling constructor

include/dlr.h
samskalicky: do we make any claims about API binary compatibility between DLR versions? or do we assume users would recompile their app with the latest dlr.h header file to use a newer version of DLR?

include/dlr.h
```diff
@@ -47,6 +47,17 @@ typedef void (*DLRFreeFunctionPtr)(void*);
 typedef void* (*DLRMemalignFunctionPtr)(size_t, size_t);
 #endif
 
+#ifndef DLR_MODEL_ELEM
+#define DLR_MODEL_ELEM
+enum DLRModelElemType { HEXAGON_LIB, NEO_METADATA, TVM_GRAPH, TVM_LIB, TVM_PARAMS, RELAY_EXEC };
```
trevor-m, line 52: I agree with sam that a string type would be easier to maintain/extend in future.

include/dlr.h
apivovarov: It should be possible to use new libdlr.so and older dlr.h in user's applications. Enum is just an integer number. We will add new element to the end of the list.

include/dlr.h
apivovarov: Well, if we support new elem type we will need to document it in the CreateDLRModel function documentation. Anyway it will be code change in dlr.h, in cc files and in unit tests.
Enum is self-documented.
Inside the users application it should be possible to use new libdlr.so (included to the model) and older dlr.h. It should work in case they do not use newely added types in the application. If user want to use new type - they need to update dlr.h. But it's better to do it anyway - in order to get the documentation of the new type.

src/dlr_relayvm.cc
apivovarov: fixed

src/dlr_relayvm.cc
apivovarov: fixed

include/dlr.h
samskalicky: why would it be a code change in dlr.h if we add a new file type string?

trevor-m: Ok that makes sense, thanks.

src/dlr_relayvm.cc
```diff
@@ -48,10 +59,42 @@ void RelayVMModel::SetupVMModule() {
...
+    } else if (el.type == DLRModelElemType::TVM_LIB) {
+      if (el.path != nullptr) {
+        model_lib_path = el.path;
+      } else {
+        throw dmlc::Error("Invalid RelayVM model element TVM_LIB");
+      }
+    } else if (el.type == DLRModelElemType::NEO_METADATA) {
+      if (el.path != nullptr) {
+        metadata_path = el.path;
+      } else {
```
samskalicky, line 85: metadata can be passed directly as a string too, not just as a path

include/dlr.h
```diff
@@ -47,6 +47,17 @@ typedef void (*DLRFreeFunctionPtr)(void*);
...
+typedef struct ModelElem {
+  const DLRModelElemType type;
```
samskalicky, line 54: can we add constructors for the struct to simplify setting just a path, or data & size?

src/dlr_tvm.cc
```diff
@@ -57,27 +69,68 @@ void TVMModel::SetupTVMModule(std::vector<std::string> model_path) {
...
+    } else if (el.type == DLRModelElemType::TVM_LIB) {
+      if (el.path != nullptr) {
+        model_lib_path = el.path;
+      } else {
+        throw dmlc::Error("Invalid TVM model element TVM_LIB");
+      }
+    } else if (el.type == DLRModelElemType::NEO_METADATA) {
+      if (el.path != nullptr) {
+        metadata_path = el.path;
+      } else {
```
samskalicky, line 114: same comment here, metadata could be set as data or a path

src/dlr_relayvm.cc
apivovarov: Added that.

src/dlr_relayvm.cc
```diff
@@ -48,10 +59,53 @@ void RelayVMModel::SetupVMModule() {
...
+        "Invalid RelayVM model. Must have RELAY_EXEC, TVM_LIB and NEO_METADATA elements");
+  }
+
+  // Load Metadata from file or String
+  if (!metadata_path.empty() && !IsFileEmpty(metadata_path)) {
+    LoadJsonFromFile(metadata_path, this->metadata_);
+    ValidateDeviceTypeIfExists();
+  } else if (!metadata_data.empty()) {
+    LoadJsonFromString(metadata_data, this->metadata_);
+    ValidateDeviceTypeIfExists();
```
samskalicky, line 105: should we just do this up above? or just load the file from path into a string above, and always call LoadJsonFromString here?

include/dlr.h
apivovarov: Looks like in C they just use `{a,b,c}` to create and init a struct.

src/dlr_tvm.cc
apivovarov: fixed

src/dlr_relayvm.cc
apivovarov: Removed `metadata_path`, added helper function `dlr::LoadFileToString`

samskalicky (approved):
Thanks @apivovarov !

src/dlr_relayvm.cc
```diff
@@ -48,10 +59,45 @@ void RelayVMModel::SetupVMModule() {
...
-                        std::istreambuf_iterator<char>());
+  std::string code_data;
+  std::string model_lib_path;
+  std::string metadata_data;
+  for (DLRModelElem el : model_elems) {
+    if (el.type == DLRModelElemType::RELAY_EXEC) {
+      if (el.path != nullptr) {
+        std::ifstream relay_ob(el.path, std::ios::binary);
+        code_data.assign((std::istreambuf_iterator<char>(relay_ob)),
+                         std::istreambuf_iterator<char>());
```
samskalicky, line 70: should we use your new `LoadFileToString` here too?

src/dlr_common.cc
```diff
@@ -68,7 +68,23 @@ void dlr::LoadJsonFromFile(const std::string& path, nlohmann::json& jsonObject)
...
+  } catch (nlohmann::json::exception&) {
+    jsonObject = nullptr;
+  }
+}
+
+std::string dlr::LoadFileToString(const std::string& path, std::ios_base::openmode mode) {
+  std::ifstream fstream(path, mode);
+  std::stringstream blob;
+  blob << fstream.rdbuf();
+  return blob.str();
```
samskalicky, line 84: @trevor-m whats the diff between doing this ^^ versus:
```
std::ifstream relay_ob(el.path, std::ios::binary);
        code_data.assign((std::istreambuf_iterator<char>(relay_ob)),
                         std::istreambuf_iterator<char>());
```
that you suggested?

src/dlr_relayvm.cc
apivovarov: Probably. Lets check why dlr_relayvm.cc loads binary files to string this way. I guess we can do "the best" approach to load files in LoadFileToString in the next PR.

src/dlr_common.cc
trevor-m: I'm not sure, it seems like it might be the same

apivovarov: ok, using LoadFileToString for relay_exec ro files as well.

src/dlr_tvm.cc
```diff
@@ -57,27 +69,66 @@ void TVMModel::SetupTVMModule(std::vector<std::string> model_path) {
...
+        params_data = static_cast<const char*>(el.data);
+        params_size = el.data_size;
+      } else {
+        throw dmlc::Error("Invalid TVM model element TVM_PARAMS");
+      }
+    } else if (el.type == DLRModelElemType::TVM_LIB) {
+      if (el.path != nullptr) {
+        model_lib_path = el.path;
+      } else {
+        throw dmlc::Error("Invalid TVM model element TVM_LIB");
```
trevor-m, line 105: Maybe also say that TVM_LIB must be a file (cannot be from memory)

tests/cpp/dlr_tvm_elem_test.cc
```diff
@@ -0,0 +1,69 @@
...
+        {DLRModelElemType::NEO_METADATA, nullptr, meta_str.c_str(), 0}};
+
+    // Setup input data
+    img = LoadImageAndPreprocess("cat224-3.txt", img_size, batch_size);
+
+    // Instantiate model
+    int device_type = 1;
+    int device_id = 0;
+    DLContext ctx = {static_cast<DLDeviceType>(device_type), device_id};
+    model = new dlr::TVMModel(model_elems, ctx);
```
apivovarov, line 39: We should also add tests for the case when TVM_LIB is given by mistake as a pinter in memory instead of a file. Plus the test case when required elements are missing

apivovarov: Added more tests

src/dlr_tvm.cc
apivovarov: fixed

trevor-m (approved):
Thanks @apivovarov !

PR #303, opened 2020-12-02
Add SparseInput flag to metadata for treelite models

trevor-m:
While https://github.com/neo-ai/neo-ai-dlr/pull/297 was correct for some Treelite models which expect sparse inputs, it produced incorrect results for Treelite models expecting dense inputs (where zero is different than missing).

With this PR, the change from #297 is only enabled if a "SparseInput" flag is set in the metadata.

"Sparse" elements are treated as "missing" by the tree booster and as zeros by the linear booster.
https://github.com/dmlc/xgboost/issues/1634
https://www.kaggle.com/c/porto-seguro-safe-driver-prediction/discussion/40835


trevor-m:
https://github.com/neo-ai/neo-ai-dlr/pull/321

Issue #306, opened 2020-12-08
Prepare Universal DLR python wheel


apivovarov:
For many targetDevices `libdlr.so` is included into the model artifact produced by Sagemager Neo (e.g. RPi)
However, DLR PyPI package is not available for many platforms.
In that case in order to run the model using Python API customers still need to build DLR package from source. But building C/C++ code is not actually necessary since they already have pre-built `libdlr.so` in the model artifact.
Customers should be able to easily install "universal" version of DLR package using pip. That "generic" package can have only python code.
We need to investigate what should be uploaded to PyPI repo in order to create this "universal" DLR package on the client side.
Maybe it should be a source version of DLR pkg with build script which skips C/C++ code building.

apivovarov:
Universal Wheels:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

https://cheat.readthedocs.io/en/latest/python/packaging.html

https://stackoverflow.com/questions/30438216/how-do-i-upload-a-universal-python-wheel-for-python-2-and-3

PR #308, opened 2020-12-10
Allow option to build python-only wheel

trevor-m:
Since Neo now will include the libdlr.so with the compiled model itself, most users won't actually need to build their own libdlr.so. This PR adds option to build a python-only DLR wheel which doesn't have libdlr.so, so the user is not required to build anything from source.

Thanks @apivovarov for the idea
https://github.com/neo-ai/neo-ai-dlr/issues/306

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -1,26 +1,32 @@
 import os
 import io
+import sys
 from setuptools import setup, find_packages
 from subprocess import check_output
 from setuptools.dist import Distribution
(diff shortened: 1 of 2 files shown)


apivovarov:
ok, ci_build will have `dlr-1.6.0-py3-none-manylinux1_x86_64.whl` (py2 was removed)

apivovarov (approved):

PR #309, opened 2020-12-10
Use plat-name in create_wheel.sh

apivovarov:
Use setup.py `--plat-name` option to create wheel file with required suffix.
e.g.
```
# command
tests/ci_build/create_wheel.sh manylinux1_x86_64
# creates
python/dist/dlr-1.6.0-py3-none-manylinux1_x86_64.whl
```


trevor-m (approved):
Thanks @apivovarov

PR #325, opened 2021-02-12
Add DataTransform for DateTime

yunma10:

### DIFF
--- a/include/dlr_data_transform.h
+++ b/include/dlr_data_transform.h
@@ -95,7 +95,8 @@ class DLR_DLL DateTimeTransformer : public Transformer {
 /*! \brief Handles transformations of input and output data. */
 class DLR_DLL DataTransform {
  private:
-  /*! \brief When there is no mapping entry for TransformOutput, this value is used. */
+  /*! \brief When there is no mapping entry for TransformOutput, this value is
+   * used. */
   const char* kUnknownLabel = "<unseen_label>";
 
   /*! \brief Buffers to store transformed outputs. Maps output index to
--- a/include/dlr_data_transform.h
+++ b/include/dlr_data_transform.h
@@ -14,14 +14,12 @@ namespace dlr {
 /*! \brief Base case for input transformers. */
 class DLR_DLL Transformer {
  public:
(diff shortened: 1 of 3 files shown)


src/dlr_data_transform.cc
```diff
@@ -126,28 +142,149 @@ void CategoricalStringTransformer::MapToNDArray(const nlohmann::json& input_json
...
+  std::string date_str = GetNextSplittedStr(input_string, delimiter);
+  std::string year_str = GetNextSplittedStr(input_string, delimiter);
+  std::string time_str = GetNextSplittedStr(input_string, delimiter);
+
+  delimiter = " ";
+  std::string month_str = GetNextSplittedStr(date_str, delimiter);
+  std::string day_str = GetNextSplittedStr(date_str, delimiter);
+
+  delimiter = ":";
+  bool isPM = time_str.substr(time_str.size() - 2, 2).compare("pm") == 0;
```
trevor-m, line 206: Nit: rename to `is_pm`

include/dlr_data_transform.h
```diff
@@ -7,61 +7,120 @@
 
 #include "dlr_common.h"
 
+#define NUM_DATE_TIME_COLS 7
```
trevor-m, line 10: We could make this a private const member variable of `DateTimeTransformer` to be consistent with other constant values in other transformers.
`const int kNumDateTimeCols = 7;`

include/dlr_data_transform.h
```diff
@@ -7,61 +7,120 @@
...
+  void MapToNDArray(const nlohmann::json& input_json,
+                    const nlohmann::json& transform,
+                    tvm::runtime::NDArray& input_array) const;
+};
+
+class DLR_DLL DateTimeTransformer : public Transformer {
+ private:
+  /*! \brief When there is no mapping entry for TransformInput, this value is
+   * used. */
   const float kMissingValue = -1.0f;
```
trevor-m, line 56: I think kMissingValue is not used for DateTimeTransformer so you can remove.

trevor-m (changes requested):
Thanks @yunma10 

Just some minor recommendations. Please also try to fix the clang formatting.

PR #330, opened 2021-02-25
Enable TVM Runtime's `TVM_C_API_MEM_MGMT` flag to fix mem leak

mycpuorg:
https://github.com/neo-ai/tvm/pull/187 fixes an important
memory leak that can be avoided in C/C++ clients.
DLR is one such client.

Thanks for contributing to DLR! By submitting this pull request, you confirm that your contribution is made under the terms of the [Apache 2.0 license](https://github.com/neo-ai/neo-ai-dlr/blob/main/LICENSE).

Please refer to our [guideline](https://github.com/neo-ai/neo-ai-dlr/blob/main/CONTRIBUTING.md) for useful information and tips.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,9 @@ set(DMLC_CORE_SRC "${TVM_SRC}/3rdparty/dmlc-core")
 set(DLPACK_SRC "${TVM_SRC}/3rdparty/dlpack")
 set(FMT_SRC "${TREELITE_SRC}/3rdparty/fmt")
 
+# DLR only uses C-APIs from TVM RunTime
+set(TVM_C_API_MEM_MGMT ON)
+
 include_directories("${TVM_SRC}/include")
 include_directories("${TVM_SRC}/src/runtime")
 include_directories("${DLPACK_SRC}/include")


jiajuns (approved):

mycpuorg:
@samskalicky @trevor-m Can you please bless this?

samskalicky:
I dont think we should set TVM_C_API_MEM_MGMT to ON by default in DLR. We specifically benefit from the static allocation in some use-cases where we had issues during shutdown with segfaults because some libraries are cleaned up before others. 

I suggest we make this an option in DLR and leave it to OFF by default, and you can set it on for your use case when you build DLR.

trevor-m:
> I dont think we should set TVM_C_API_MEM_MGMT to ON by default in DLR. We specifically benefit from the [static allocation in TVM registry](https://github.com/jiajuns/tvm/blob/release-1.8.0-rc/src/runtime/registry.cc#L54-L55) in some use-cases where we had issues during shutdown with segfaults because some libraries are cleaned up before others. Heres the original thread: [apache/tvm#5986](https://github.com/apache/tvm/pull/5986)
> 
> I suggest we make this an option in DLR and leave it to OFF by default, and you can set it on for your use case when you build DLR.

Agree

jiajuns:
> I dont think we should set TVM_C_API_MEM_MGMT to ON by default in DLR. We specifically benefit from the [static allocation in TVM registry](https://github.com/jiajuns/tvm/blob/release-1.8.0-rc/src/runtime/registry.cc#L54-L55) in some use-cases where we had issues during shutdown with segfaults because some libraries are cleaned up before others. Heres the original thread: [apache/tvm#5986](https://github.com/apache/tvm/pull/5986)
> 
> I suggest we make this an option in DLR and leave it to OFF by default, and you can set it on for your use case when you build DLR.

As mentioned in the original thread, that it is recommended to 

> _"allocate raw pointer for the device API and never destory themselves(as the resource will de-allocate in unloading and no de-allocation is needed)."_

However what we found in our testing is that `dlclose()` will not deallocate those objects. Therefore as a result, memory will creep up during runtime. As far as I know, this race to clean up will be a problem for other language rather than C. Do we know which part of Neo DLR might be affected?

samskalicky:
> However what we found in our testing is that `dlclose()` will not deallocate those objects. 

Its not `dlclose()` that will deallocate the object, its the OS cleanup of the process when the process exits. 

> Therefore as a result, memory will creep up during runtime. 

Memory will not "creep" since this is a static allocation that only happens once in the process in typical use cases where a user loads their model/DLR and runs inference. If you are testing an exotic use case where you open/close DLR a bunch of times, then it may be an issue for you. In which case it makes sense to make this option specific to your use case.

apivovarov:
I do not think this change is enough to add -DTVM_C_API_MEM_MGMT flag to gcc
See my comment in TVM PR [here](https://github.com/neo-ai/tvm/pull/187)

btw, you can `make VERBOSE=1` and see that this change does nothing to `src/runtime/registry.cc` compilation.

PR #333, opened 2021-03-12
Add a CMAKE option to enable/disable datatransform; default OFF

yunma10:
Strptime function is specific to LINUX system. This CMAKE option ensures that DataTransform functions for Sagemaker-Extension-Scikit-Learn models are Turned off by default (This feature will only be used on LINUX machines). 

To enable:
 cmake -DENABLE_DATATRANSFORM=ON ..


trevor-m (approved):
Thanks @yunma10 !

I think it would be nice to add some code to give an error message on model load if DataTransform is present in metadata but wasn't enabled during compilation. We can add it later though.

PR #336, opened 2021-03-19
add datetime templates, fix bug in week number calculation

yunma10:
-add datetime template for "%Y-%m-%d %H:%M:%S+00", "%H:%M:%S+00", "%H:%M:%S"
-fix undefined behavior when YY/MM/DD is undefined
-fix bug in week number calculation

### DIFF
--- a/src/dlr_data_transform.cc
+++ b/src/dlr_data_transform.cc
@@ -140,33 +140,53 @@ tvm::runtime::NDArray DateTimeTransformer::InitNDArray(const nlohmann::json& inp
   return tvm::runtime::NDArray::Empty(arr_shape, dtype, ctx);
 }
 
-int64_t DateTimeTransformer::GetWeekDay(int64_t year, int64_t month, int64_t day) const {
-  int64_t century = year / 100;
-  int64_t month_ = month == 2 ? 12 : postive_modulo((month - 2), 12);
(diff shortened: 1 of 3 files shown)


src/dlr_data_transform.cc
```diff
@@ -140,41 +140,61 @@ tvm::runtime::NDArray DateTimeTransformer::InitNDArray(const nlohmann::json& inp
...
 void DateTimeTransformer::DigitizeDateTime(std::string& input_string,
                                            std::vector<int64_t>& datetime_digits) const {
-  struct tm tm = {};
+  struct tm tm {};
 
   char* strptime_success;
   for (const auto datetime_template : datetime_templates) {
     strptime_success = strptime(input_string.c_str(), datetime_template.c_str(), &tm);
     if (strptime_success) {
+      if (datetime_template.compare(0, 8, "%H:%M:%S") == 0) {
```
trevor-m, line 182: Could you add a test entry for this case?

trevor-m (changes requested):

src/dlr_data_transform.cc
yunma10: Just added a test to test only fields of hour/minute/second. Other fields are dependent on the current local time.

trevor-m (approved):

PR #340, opened 2021-04-01
avoid targeting android with -lpthread flag

liangfu:
On some old version of CMake, it consider the expression `ANDROID_BUILD` in `else(ANDROID_BUILD)` as a condition, therefore, the statement is misleading. Removing expression in `else()` statement would provide a clear logic to CMake.


liangfu:
cc @trevor-m

trevor-m (approved):

PR #341, opened 2021-04-01
Maintain order of input names to fix wrong dtype issue

trevor-m:
The order of the input names was being lost because it was sorted alphabetically when removing weight names. This led to DLR matching the input to the wrong index and therefore wrong dtype.

This probably wasn't an issue before because most models only have one input or all float32 inputs.

### Before

name, index, dtype
```
image 1 float32                                             
transform 3 int32
ValueError: input data with name transform should have dtype int32 but float32 is provided
```

### After

name, index, dtype
```
image 0 float32
transform 1 float32
bbox 2 float32
index 3 int32
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -471,6 +471,22 @@ if(NOT(AAR_BUILD))
     file(REMOVE /tmp/${INVERSELABEL_MODEL})
   endif()
 
+  set(INPUTORDER_MODEL input_order-ml_m4.tar.gz)
+  set(INPUTORDER_MODEL_DIR ${CMAKE_CURRENT_BINARY_DIR}/input_order)
+  if(NOT IS_DIRECTORY ${INPUTORDER_MODEL_DIR})
+    file(MAKE_DIRECTORY ${INPUTORDER_MODEL_DIR})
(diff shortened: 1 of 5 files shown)


trevor-m:
Unit test failure without this PR
```
[ RUN      ] DLR.TestDLRInputOrder
/data/neo-ai-dlr/tests/cpp/dlr_test.cc:401: Failure
Expected equality of these values:
  input_names[0]
    Which is: "bbox"
  "image"
/data/neo-ai-dlr/tests/cpp/dlr_test.cc:402: Failure
Expected equality of these values:
  input_names[1]
    Which is: "image"
  "transform"
/data/neo-ai-dlr/tests/cpp/dlr_test.cc:403: Failure
Expected equality of these values:
  input_names[2]
    Which is: "index"
  "bbox"
/data/neo-ai-dlr/tests/cpp/dlr_test.cc:404: Failure
Expected equality of these values:
  input_names[3]
    Which is: "transform"
  "index"
[  FAILED  ] DLR.TestDLRInputOrder (0 ms)
```

apivovarov (approved):

PR #349, opened 2021-04-27
Fix dlr_pipeline_test for ASAN

apivovarov:
One of the dlr_pipeline_test test cases does not delete DLR model at the end.
It causes AddressSanitizer errors
```
Direct leak of 200 byte(s) in 1 object(s) allocated from
...
Indirect leak of 2208 byte(s) in 3 object(s) allocated from:
...
```

To run `dlr_pipeline_test` with AddressSanitizer enabled use the following flags:
```
cmake .. \
-DCMAKE_CXX_FLAGS="-fsanitize=address -g3" \
-DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address -g3" \
-DCMAKE_BUILD_TYPE=Debug

make -j16

./dlr_pipeline_test
```

to install AddressSanitizer on Ubuntu
```
apt install libasanN

# Select libasan N version depending on your gcc:
# libasan2: gcc-5
# libasan3: gcc-6
# libasan4: gcc-7
# libasan5: gcc-8
```

### DIFF
--- a/include/dlr_treelite.h
+++ b/include/dlr_treelite.h
@@ -22,7 +22,8 @@ struct TreeliteInput {
   std::vector<size_t, DLRAllocator<size_t>> row_ptr;
   size_t num_row;
   size_t num_col;
-  CSRBatchHandle handle;
+  CSRBatchHandle handle = nullptr;
+  ~TreeliteInput();
 };
 
 /*! \brief Get the paths of the Treelite model files.
--- a/demo/cpp/model_peeker.cc
+++ b/demo/cpp/model_peeker.cc
@@ -145,5 +145,7 @@ int main(int argc, char** argv) {
 
   peek_model(model);
 
+  // cleanup
+  DeleteDLRModel(&model);
   return 0;
 }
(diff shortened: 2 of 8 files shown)


trevor-m (approved):
LGTM

include/dlr_treelite.h
```diff
@@ -23,6 +23,11 @@ struct TreeliteInput {
   size_t num_row;
   size_t num_col;
   CSRBatchHandle handle;
+
+  ~TreeliteInput() {
+    if (handle) TreeliteDeleteSparseBatch(handle);
```
trevor-m, line 28: Does handle need to be initialized to nullptr in constructor?

apivovarov: fixed. Looks like in C++11 we can just say `CSRBatchHandle handle = nullptr;`

yunma10 (approved):

trevor-m:
Thanks @apivovarov 
Could you also open a PR for main branch later?

PR #357, opened 2021-05-25
Fix pthread linking issue for AAR_BUILD

apivovarov:
Android NDK does not have pthread library. We should exclude linking to pthread when building AAR file.

### Tests
- AAR build - OK
- libdlr.so for Android - OK
- demo apps for Linux - OK (gcc)
- demo apps for Android - Failed. clang++ complains about  3rdparty/libnpy/npy.hpp lines [66](https://github.com/llohse/libnpy/blob/641f2a581357a51c0973889f031ad88db28de31a/npy.hpp#L66) and 67
More on the issue: https://stackoverflow.com/questions/46576847/clang-vs-gcc-crtp-constexpr-variable-cannot-have-non-literal-type


zxy844288792 (approved):

trevor-m (approved):
Thanks @apivovarov

apivovarov:
npy.hpp issue was fixed here https://github.com/llohse/libnpy/commit/3e0e373828d88b0a1df2ba73cc31616a848b1aef
We need to upgrade it to the latest version

PR #358, opened 2021-05-25
Update libnpy

apivovarov:
Update libnpy to the latest version. It fixes "[clang] constexpr variable cannot have non-literal type" https://github.com/llohse/libnpy/issues/14


trevor-m (approved):
Thanks @apivovarov !

PR #362, opened 2021-06-17
Add USE_MLAS, USE_OPENMP build flags

trevor-m:
Thanks for contributing to DLR! By submitting this pull request, you confirm that your contribution is made under the terms of the [Apache 2.0 license](https://github.com/neo-ai/neo-ai-dlr/blob/main/LICENSE).

Please refer to our [guideline](https://github.com/neo-ai/neo-ai-dlr/blob/main/CONTRIBUTING.md) for useful information and tips.


CMakeLists.txt
```diff
@@ -58,6 +58,8 @@ option(USE_CUDA  "Build with CUDA" OFF)
 option(USE_CUDNN "Build with CUDNN" OFF)
 option(USE_TENSORRT "Build with Tensor RT" OFF)
 option(ENABLE_DATATRANSFORM "Enable datatransform for Sagemaker-scikit-learn-extension models" OFF)
+option(USE_OPENMP "Build with OpenMP" OFF)
```
ymwangg, line 61: It looks like `option` cannot take strings other than `ON` and `OFF`, should we use `set` here like https://github.com/neo-ai/tvm/blob/dev/cmake/config.cmake#L151.

ymwangg:
A few comments:
1. Should this be merged into branch `release-1.9.1`? I didn't see the main branch has a recent mlas-enabled tvm version.
2. It looks like this change does not build as expected, could you double check?

CMakeLists.txt
trevor-m: I was able to use `cmake .. -DUSE_MLAS=ON -DUSE_OPENMP=intel`

trevor-m:
> A few comments:
> 
>     1. Should this be merged into branch `release-1.9.1`? I didn't see the main branch has a recent mlas-enabled tvm version.
> 
>     2. It looks like this change does not build as expected, could you double check?

Thanks for the review @ymwangg! That is true, this should be merged into release-1.10.0-rc. Then it can be cherry-picked to 1.9.1. It will make it's way into main branch when 1.10 is finished and it is merged to main.

CMakeLists.txt
ymwangg: I see, that's overriding the `option` with `set` through command line.

trevor-m: I see, I think this is ok since it is consisent with other DLR flags such as USE_OPENCL, USE_TENSORRT which are defined with `option` but can be overriding with `set` from command line.

PR #364, opened 2021-08-04
Update: bring new commits from TVM 1.10.0-rc

jsheng-jian:
Update: bring new commits from TVM 1.10.0-rc
#363


trevor-m (approved):
Thanks Jian!

PR #365, opened 2021-08-17
Add support and tests for score output using variable pred_margin

yunma10:
User can enable the score output in replacement of probabilities by using the function SetPredMargin().

### DIFF
--- a/include/dlr_treelite.h
+++ b/include/dlr_treelite.h
@@ -84,7 +84,7 @@ class DLR_DLL TreeliteModel : public DLRModel {
   virtual void SetNumThreads(int threads) override;
   virtual void UseCPUAffinity(bool use) override;
 
-  inline void SetPredMargin() { pred_margin = 1; };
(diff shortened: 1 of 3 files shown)


include/dlr_treelite.h
```diff
@@ -80,6 +83,8 @@ class DLR_DLL TreeliteModel : public DLRModel {
   virtual void Run() override;
   virtual void SetNumThreads(int threads) override;
   virtual void UseCPUAffinity(bool use) override;
+
+  inline void SetPredMargin() { pred_margin = 1; };
```
trevor-m, line 87: I think we should allow user to enable and disable, could you add a boolean arg to `SetPredMargin()`?

trevor-m (changes requested):

include/dlr_treelite.h
yunma10: Will do

trevor-m (approved):

PR #369, opened 2021-09-28
[DataTransform] Add TextDataTransform and miscellaneous Modifications

CloudManX:
DateTimeTransformer: Delete Unused function and change templates from vector to array since there is no need for growing.
PredMargin: Eliminate Ambiguity by adding *this
TextTransformer: 

- Addition of a new transformer to handle inputs for MultiColumnTfidfVectorizer
- Mimic the behavior of a CountVectorizer for each Column([(https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html)])

### DIFF
--- a/include/dlr_data_transform.h
+++ b/include/dlr_data_transform.h
@@ -17,15 +17,13 @@ class DLR_DLL Transformer {
   const tvm::runtime::NDArray empty_;
 
  public:
-  virtual void MapToNDArray(const nlohmann::json& input_json,
-                            const nlohmann::json& transform,
+  virtual void MapToNDArray(const nlohmann::json& input_json, const nlohmann::json& transform,
                             tvm::runtime::NDArray& input_array) const = 0;
 
   /*! \brief Helper function for TransformInput. Allocates NDArray to store
    * mapped input data. */
-  virtual void InitNDArray(const nlohmann::json& input_json,
-                           const nlohmann::json& transform, DLDataType dtype,
-                           DLContext ctx,
+  virtual void InitNDArray(const nlohmann::json& input_json, const nlohmann::json& transform,
+                           DLDataType dtype, DLContext ctx,
                            tvm::runtime::NDArray& input_array) const;
 };
 
@@ -36,8 +34,7 @@ class DLR_DLL FloatTransformer : public Transformer {
   const float kBadValue = std::numeric_limits<float>::quiet_NaN();
 
  public:
-  void MapToNDArray(const nlohmann::json& input_json,
-                    const nlohmann::json& transform,
+  void MapToNDArray(const nlohmann::json& input_json, const nlohmann::json& transform,
                     tvm::runtime::NDArray& input_array) const;
 };
 
@@ -48,8 +45,7 @@ class DLR_DLL CategoricalStringTransformer : public Transformer {
   const float kMissingValue = -1.0f;
 
  public:
-  void MapToNDArray(const nlohmann::json& input_json,
-                    const nlohmann::json& transform,
+  void MapToNDArray(const nlohmann::json& input_json, const nlohmann::json& transform,
                     tvm::runtime::NDArray& input_array) const;
 };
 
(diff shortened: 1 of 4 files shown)


include/dlr_data_transform.h
```diff
@@ -80,16 +82,46 @@ class DLR_DLL DateTimeTransformer : public Transformer {
...
+  virtual void InitNDArray(const nlohmann::json& input_json, const nlohmann::json& transform,
+                           DLDataType dtype, DLContext ctx,
+                           tvm::runtime::NDArray& input_array) const override;
+
+  inline void SetIndex(int idx) const { column_idx = idx; };
+
+ private:
+  const static int kCharNum = 256;
+  char delims[kCharNum];
+  std::unique_ptr<std::vector<std::unordered_map<std::string, int>>> vocab_to_cols;
```
trevor-m, line 101: Let's rename class member variables to end with `_` to follow our conventions, for example: `vocab_to_cols_`

src/dlr_data_transform.cc
```diff
@@ -245,6 +238,73 @@ nlohmann::json DataTransform::TransformOutputHelper1D(const nlohmann::json& tran
...
+  int id = col_to_id->at(column_idx);
+  std::unordered_map<std::string, int>& vocab_to_col = vocab_to_cols->at(id);
+  int num_col = vocab_to_col.size();
+  float* data = static_cast<float*>(input_tensor->data);
+
+  for (size_t r = 0; r < input_json.size(); ++r) {
+    memset(data + r * num_col, 0, num_col * sizeof(float));
+    std::string entry = input_json[r][column_idx].get_ref<const std::string&>();
+    char* pch;
+    pch = strtok(&entry[0], delims);
```
trevor-m, line 295: We shouldn't use C library functions like `strtok` since it is considered unsafe. Can you refactor this to use C++ functions like `std::string::find_first_of` or stringstream?

src/dlr_data_transform.cc
```diff
@@ -245,6 +238,73 @@ nlohmann::json DataTransform::TransformOutputHelper1D(const nlohmann::json& tran
...
+    while (pch != NULL) {
+      std::string token(pch);
+      LowerStr(token);
```
trevor-m, line 298: can lowerstr be done only once on the whole string instead?

trevor-m (changes requested):
Thanks @CloudManX

src/dlr_data_transform.cc
```diff
@@ -245,6 +238,73 @@ nlohmann::json DataTransform::TransformOutputHelper1D(const nlohmann::json& tran
...
+  CHECK_EQ(input_tensor->ctx.device_type, DLDeviceType::kDLCPU)
+      << "DataTransform TfIdfVectorizer is only supported for CPU.";
+
+  int id = col_to_id->at(column_idx);
+  std::unordered_map<std::string, int>& vocab_to_col = vocab_to_cols->at(id);
+  int num_col = vocab_to_col.size();
+  float* data = static_cast<float*>(input_tensor->data);
+
+  for (size_t r = 0; r < input_json.size(); ++r) {
+    memset(data + r * num_col, 0, num_col * sizeof(float));
```
trevor-m, line 292: `std::fill_n(data + r * num_col, num_col, 0.0f)` is better option for zeroing out float array

src/dlr_data_transform.cc
CloudManX: Do you mean on the who input string, not after tokenization?

include/dlr_data_transform.h
CloudManX: Done.

src/dlr_data_transform.cc
CloudManX: Done

src/dlr_data_transform.cc
CloudManX: Done

trevor-m (approved):
Thanks @CloudManX

PR #371, opened 2021-10-29
Fixing excessive memory usage in texttransformer by cleaning the look…

CloudManX:
[Sklearn DataTransform]
+ Temp variable cleaning preventing excessive memory growing in RedShift container

### DIFF
--- a/src/dlr_data_transform.cc
+++ b/src/dlr_data_transform.cc
@@ -301,6 +301,8 @@ void TextTransformer::MapToNDArray(const nlohmann::json& input_json,
       }
     }
   }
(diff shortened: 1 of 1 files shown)


apivovarov:
What you think if we add some sort of a Clean method to `Transformer`? Then the sequence of `transformer` calls in `TransformInput` method will be the following

```
void DataTransform::TransformInput {
   for () {
      transformer->InitNDArray(...);
      transformer->MapToNDArray(...);
      transformer->Clean();
```

apivovarov:
@CloudManX

PR #406, opened 2022-01-30
[Treelite] multi:softmax output shape fix

apivovarov:
Treelite model output shape will be know only after the model run.
If model uses objective multi:softmax then output shape will be (batch, 1) because predictor will execute predictor_transform max_index.

mnist model uses multi:softmax objective which outputs one class with the maximum probability.

Related discussion: https://github.com/dmlc/treelite/issues/351


jsheng-jian (approved):
Merged, thanks!

PR #407, opened 2022-01-30
[Treelite] Add TreeliteModel destructor

apivovarov:
Add TreeliteModel destructor

`TreeliteModel` class has `PredictorHandle treelite_model_` member.
`typedef void* PredictorHandle;`

Looks like we have a memory leak issue because we do not delete `treelite::Predictor` to which `treelite_model_` points.
`treelite_model_` is not wrapped with `unique_pointer`. wrapping `void*` with `unique_pointer` does not work anyway - it needs actual object type such as `predictor::Predict`.

it is easier to just add explicit cleanup method to `TreeliteModel` to fix memory leak issue.


jsheng-jian (approved):
Looks good, thanks Alex!

PR #408, opened 2022-01-30
Use get-pip.py for python 3.6

apivovarov:
Our Docker images are based on Ubuntu 18.04 which install python 3.6 by default. This python is quite old. Now we have to use old get-pip.py for Python 3.6 to fix the following CI error
```
ERROR: This script does not work on Python 3.6 The minimum supported Python version is 3.7. Please use https://bootstrap.pypa.io/pip/3.6/get-pip.py 
```


jsheng-jian (approved):
Merged, thanks!

PR #428, opened 2022-04-22
[TF C] Fix GetDLRInput/Output Type to return type name

apivovarov:
Tensorflow C API:
Fix `GetDLRInputType` and `GetDLROutputType` to return type name instead of integer type_id.
Type name example: `DT_FLOAT`


jsheng-jian (approved):
Thanks Alex, approved!

